PR #11, opened 2019-02-18
Update devnet docz

evgenykuzyakov:
Fixes  https://github.com/nearprotocol/nearcore/issues/581


evgenykuzyakov:
Seems this broke docs.

PR #14, opened 2019-02-22
pynear update

azban:


azban:
required for https://github.com/nearprotocol/nearcore/pull/603

azban:
this needs to go out, since we updated nearcore

PR #19, opened 2019-03-16
Update the client lib with examples

potatodepaulo:
Added some examples to the nearlib docs. 
These need fleshing out, but it's better than nothing.
Early next week, I'd like to go through some more realistic examples with whoever has time. 
E.g. WalletAccount use cases that might appear in the wild that are a little more meaningful

### DIFF
--- a/docs/client-api/js.md
+++ b/docs/client-api/js.md
@@ -1,6 +1,5 @@
 ---
 name: NEARLib.js
-route: /lib/js
 menu: Client API
 ---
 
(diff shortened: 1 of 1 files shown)


vgrichina:
@potatodepaulo this file is auto-generated, you don't want to update it directly.

potatodepaulo:
> @potatodepaulo this file is auto-generated, you don't want to update it directly.

Took a look. This one is actually a manual copy of API.md which is the true auto-generated file inside of nearcore. We're going to need to build the pipes that move the generated docs to this repo.

PR #30, opened 2019-04-23
zero to hero tutorial

ckshei:
Open to any and all feedback. Some of my thoughts: 
- I'd probably want to change the description of my tutorial to something a bit more clear: (maybe something like ramp up quickly on the Near Protocol by building an Inbound Data Oracle + a Bet Contract). 
- Might be nice to walk through how to use the fiddle, although I tried to make it quite self-explanatory
- Ideally, updating the gulp file does not have to be part of the tutorial. 
- I know there's an existing "Oracle Tutorial" that has yet to be completed, so I just called this a "Zero to Hero" tutorial because it's a great tutorial for complete beginners, and it goes over most of the knowledge necessary to build on Near. Wallet integration would probably be the final piece I would add if I had the time.


ckshei:
Awesome - thanks!

PR #33, opened 2019-05-06
Fix up globalStorage -> storage rename

vgrichina:


vgrichina:
@evgenykuzyakov let's be careful with any upcoming renames as well ;)

PR #78, opened 2019-09-27
Windows testnet

AnaisUrlichs:


potatodepaulo:
Wow, this is a total pain for windows users...

PR #98, opened 2019-10-24
add copy button to all code blocks

amgando:
found a simple copy-to-clipboard feature [on one of the sample sites](https://fbflipper.com/docs/getting-started.html) and [grabbed it from there](https://github.com/search?l=&q=clipboard+repo%3Afacebook%2Fflipper+path%3A%2Fwebsite&type=Code)

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -50,7 +50,11 @@ const siteConfig = {
   },
 
   // Add custom scripts here that would be placed in <script> tags.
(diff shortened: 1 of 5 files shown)


amgando:
actually the changes required to add a copy button is linked [right from their docs](https://docusaurus.io/docs/en/doc-markdown#adding-copy-code-buttons) to [this gist](https://gist.github.com/yangshun/55db997ed0f8f4e6527571fc3bee4675)

PR #99, opened 2019-10-24
add all possible syntax highlight colors. pick one

amgando:
higher contrast code style makes it easier for me to read.  i imagine others as well.

here's the [full list](https://github.com/highlightjs/highlight.js/tree/master/src/styles), also included in our `siteConfig.js` now too in case we want to quickly switch


potatodepaulo:
Radical!

PR #110, opened 2019-11-04
change near-shell

icerove:
change near-shell docs to latest version

### DIFF
--- a/docs/development/near-clitool.md
+++ b/docs/development/near-clitool.md
@@ -25,13 +25,13 @@ near <command>
 
 For account:
 ```bash
-  near login                                   # create a developer account
-  near create_account <accountId> <masterAccount>             # create a developer account with masterAccount, publicKey and initialBalance
-  near view <accountId>                        # view account state
-  near keys <accountId>                        # view account public keys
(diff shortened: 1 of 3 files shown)


docs/development/near-clitool.md
```diff
@@ -23,23 +23,31 @@ near <command>
 
 ### Commands
 
+For account:
+```bash
+  near login                                   # create a developer account
+  near create_account <accountId>              # create a developer account with masterAccount, publicKey and initialBalance
```
amgando: `near-shell` errors out if you don't provide the `masterAccount` here.

```
near create_account ANY_ACCOUNT_NAME
```

produces

```
Missing required argument: masterAccount
```

i think the right thing to do is to add another required parameter to this command in `near-shell` [here](https://github.com/nearprotocol/near-shell/blob/master/bin/near-cli.js#L65) based on [these docs](http://yargs.js.org/docs/#api-demandoptionkey-msg)

@icerove

icerove: we decide not to change that this time, because we will still use the command like this in many places

PR #112, opened 2019-11-08
wip. nearlib examples page

amgando:
this isn't intended for a merge, just discussion.

### DIFF
--- a/docs/development/examples/nearlib.md
+++ b/docs/development/examples/nearlib.md
@@ -1,0 +1,331 @@
+---
+id: nearlib
+title: nearlib examples
+sidebar_label: nearlib
+---
+
+## overview
+
+`nearlib` is a JavaScript/TypeScript library for development of DApps on the NEAR platform that can be used from any client or server-side JavaScript environment.
+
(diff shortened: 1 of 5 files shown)


eriktrautman:
The trickiest bit with examples is keeping them up to date.  The best case
is if the docs pull directly from the code.  Well, the best case is also
that there is a build process which has automated tests that flag when the
example/sample-app is broken but...
Anyway, is there a reasonably good solution for this?  Otherwise we end up
with more docs debt.  I know we've talked about it before but I doubt it
was implemented.

amgando:
> _ build process which has automated tests that flag when the example/sample-app is broken_

not sure what to do about this.  will keep in mind as we make progress

vgrichina:
@amgando there is a module that can be used to automatically extract examples:
https://www.npmjs.com/package/markdown-snippet-injector

It works alright (tried out in studio), just need focused effort to instrument all the code examples (something not done before).

amgando:
> a module that can be used to automatically extract examples

thanks for this, @vgrichina.  looks like it would pollute the codebase pretty heavily, no?  and we would still need to make sure people are maintaining the snippets (which almost no one does), just like comment docs.  and this would add another pre-processing step to doc generation to extract the snippets into markdown.  not feeling good about the overhead.  am i over complicating?

amgando:
closing. superceded by #136

PR #113, opened 2019-11-08
Edited hackathon page & changes to the wallet setup

AnaisUrlichs:


docs/hackathon/startup-guide.md
```diff
@@ -1,64 +1,70 @@
...
+You might have heard of distributed computing, databases, or computer networks; all of which play a role in blockchains. 
 
-This is the quick and dirty way to do that:
+Currently, most web-services utilise a single server and a single database to process your request and provide information. This infrastructure is usually managed by an individual entity, who is treating all of their data processing like a black box. Meaning, the request goes in, something happens, and the user receives an output. 
 
-## Running official TestNet locally
+While the company may make claims, and rely on third-parties to verify those, the user will never be able to verify what happened in the black box. This system relies on trust between users and companies.
 
-By default we use Docker to run the client \(with auto-updating via watchtower to upgrade the node to the new versions automatically\).
+Near is basically building the infrastructure for a new Internet which makes it harder for giant companies to steal your data and for bad guy countries to shut it down. People have been trying to figure this stuff out with similar technologies since 2008 but it’s been slow going.
```
amgando: the tone here is a little casual. specifically: _"harder for giant companies to steal your data and for bad guy countries to shut it down"_.  not sure if this makes the docs more relatable or not

docs/hackathon/startup-guide.md
```diff
@@ -1,64 +1,70 @@
...
 
-Run:
+### Setup a local environment 
+Let's set up your development environment. 
 
-```bash
-sudo apt update
-sudo apt install -y git binutils-dev libcurl4-openssl-dev zlib1g-dev libdw-dev libiberty-dev cmake gcc g++ python docker.io protobuf-compiler
-```
+1. First, you want to install the [near shell](../development/near-clitool) in your terminal;
```
amgando: can we call it `NEAR Shell`?  in general, i'd like to start using `NEAR [A-Z]___` pattern for user facing systems like NEAR Wallet, NEAR Explorer, NEAR Studio, etc.

docs/hackathon/startup-guide.md
```diff
@@ -114,6 +122,8 @@ window.contract = await near.loadContract(config.contractName, {
 });
 ```
 
+The call to `loadContract` is actually making an object with your functions that gets assigned to the `window.contract` variable so later on you can call `window.contract.myFunction`. Note that `window` is often already in scope so you can just call `contract.myFunction`.
```
amgando: note `window` will _always_ be in scope, not just _often_, in the browser.

docs/local-setup/create-account.md
```diff
@@ -21,9 +33,19 @@ Usernames can be 5-32 characters long and contain any of the following:
...
+![Step 3: Security Code](assets/entercode.png)
+
+Once you click on the "PROTECT ACCOUNT" button, you will be forwarded to the dashboard.
+
+![Wallet](assets/finalscreen.png)
+
 ### Backup your Private Keys Manually
 
-To so click below the the phone number entry on "I choose to backup my account manually".
+Alternatively, you can choose to backup your private keys manually. To do so click below the the phone number entry on "I choose to backup my account manually".
```
amgando: in this case we're not backing up "private keys" actually, we're guaranteeing that our account can be recovered using a different (ie. not phone SMS) method which is using a seed phrase

the user will not b

website/sidebars.json
```diff
@@ -2,15 +2,15 @@
   "docs": {
     "OVERVIEW": [
       "overview/who-is-this-for",
-      "overview/what-is-near"
+      "overview/what-is-near",
+      "local-setup/create-account"
```
amgando: i would move the `create-account` section to the first item in the `QUICK START` section

amgando:
@AnaisUrlichs added a few comments

docs/hackathon/startup-guide.md
AnaisUrlichs: Good point; we have not been consistent with the naming so far

docs/hackathon/startup-guide.md
AnaisUrlichs: I will change it and provide a relatable example

PR #127, opened 2019-11-22
add hotjar for more engagement visibility

amgando:


amgando:
https://www.hotjar.com

PR #138, opened 2019-12-16
Update nearlib examples and staking examples to handle yoctoNEAR change

kcole16:


docs/development/examples/nearlib/introduction.md
```diff
@@ -235,7 +235,7 @@ If you're already familiar with these terms then feel free to skip ahead.
...
 
 - **the cost of data storage and computation on the network is measured in units of gas which is a configurable bit of network economics but is usually vanishingly small**  \
-  It costs exactly 2 gas (at the time of this writing) to "compute the creation of a new account" on the network where 1 unit of gas is equivalent to 1 attoNEAR.  For a sense of scale, with the 10 NEAR in your new account you could happily afford to create a *new account for every grain of sand on the planet* (estimates suggest about 10^18 grains of sand) and still have some NEAR tokens left over for a rainy day because 1 NEAR is 10^18 attoNEAR
+  It costs exactly 2 gas (at the time of this writing) to "compute the creation of a new account" on the network where 1 unit of gas is equivalent to 1 yoctoNEAR.  For a sense of scale, with the 10 NEAR in your new account you could happily afford to create a *new account for every grain of sand on the planet* (estimates suggest about 10^24 grains of sand) and still have some NEAR tokens left over for a rainy day because 1 NEAR is 10^24 yoctoNEAR
```
amgando: the popular estimate for grains of sand on earth is still 10^18 - 10^21 so changing it to 10^24 seems too far afield.

here are a few suggestions for rewriting this statement whose intention is to make these impossible scales relatable.  which do you like best?  i think picking one is enough.

_quantity_

> For a sense of scale, with just 2 of the 10 NEAR in your new account you could happily afford to create a *new account for every grain of sand on the planet* (actually estimates suggest about 10^21 grains of sand on earth so you could do this every day for 3 years, not just once) and still have some NEAR tokens left over for a rainy day.  Gas is measured in yoctoNEAR or yNEAR which is 10^24 times smaller than NEAR.  It's very very very small.

_mass_

> For a sense of scale, if NEAR were a unit of mass and we set 1 NEAR to be the mass of your laptop then 1 unit of gas would be about the mass of a proton.  Gas is measured in yoctoNEAR or yNEAR which is 10^24 times smaller than NEAR.  It's very very very small.

> For a sense of scale, if NEAR were a unit of mass and we set 1 NEAR to the mass of this planet earth we're all living on then 1 unit of gas would be about the mass of your best friend.  Gas is measured in yoctoNEAR or yNEAR which is 10^24 times smaller than NEAR.  It's very very very small.

_time_

> For a sense of scale, if we invited 1000 of our closest friends to a party where, according to the rules of a very cruel game, every one of us has to spend 1 _billion_ gas every second then it would take us about the same as all of recorded human history to get through 1 NEAR.  Gas is measured in yoctoNEAR or yNEAR which is 10^24 times smaller than NEAR.  It's very very very small.

docs/development/examples/nearlib/introduction.md
```diff
@@ -937,16 +942,25 @@ If you saw something unexpected, here's what may have happened ...
...
 
-That's why any noticeable transfers that change the visible balance of an account in NEAR Wallet have 15 digits or more!
+`formatNearAmount()`to go from yoctoNEAR to NEAR
+
+```
+let amount_in_near, amount_in_yocto
+
+amount_in_yocto = nearlib.utils.format.parseNearAmount(amount_in_near)
+amount_in_near = nearlib.utils.format.formatNearAmount(amount_in_yocto)
+```
```
amgando: people will start calling our unit of gas "yocto" if we write this.  that's not necessarily a bad thing but we should be deliberate about these choices.  alternatively we could write these two lines as

```
amount_in_y_near = nearlib.utils.format.parseNearAmount(amount_in_near)
amount_in_near = nearlib.utils.format.formatNearAmount(amount_in_y_near)
```

docs/development/examples/nearlib/introduction.md
```diff
@@ -996,7 +1010,7 @@ At least one key will look like this with `FunctionCall` permissions (your publi
     "nonce": 0,                               // nonce is an incrementing counter to avoid duplicate transactions
     "permission": {
       "FunctionCall": {                       // this is a FunctionCall access key
-        "allowance": "100000000",             // with about 100 million units of gas (1 one-millionth of a NEAR)
+        "allowance": "100000000000000",       // with about 100 million units of gas (1 one-millionth of a NEAR)
```
amgando: this is now  100 **trillion** (`100 000 000 000 000`) and 1 **one-trillionth** of a NEAR

docs/local-setup/local-dev-testnet.md
```diff
@@ -84,7 +84,7 @@ near create_account <yourcontractname> --masterAccount=<login> --initialBalance
 ```
 
 #### masterAccount : specifies which account's key should be used to deploy the contract. This is the account ID you specified when you created your wallet. 
-#### initialBalance : specifies how much to seed the contract balance with. The min is 10 NEAR. However, in this case the amount is measured in attoNEAR (1 NEAR is 10^18 attoNEARs). Thus, you want to put at least 1000000000000000000 attoNEAR.
+#### initialBalance : specifies how much to seed the contract balance with. The min is 10 NEAR.
```
frol: I believe that "10 NEAR" is not the min value, but rather a default one. @janedegtiareva Is my understanding correct?

docs/development/examples/nearlib/introduction.md
kcole16: +1 to the first point

 Now that I'm looking at it, isn't the example actually 1 * (10^-10) of a NEAR?

kcole16: Either way, this is pretty confusing for developers (confusing myself right now). Need to think of a good way to simplify it.

Issue #48, opened 2019-07-26
RPC documentation outdated


nearmax:
Start a local node with `myaccount` account as a validator.
Issue a transaction:
```
http post http://127.0.0.1:3030/ jsonrpc=2.0 method=query params:="[\"account/myaccount\",[]]" id="dontcare"
```

Observe error message:
```
HTTP/1.1 200 OK
content-length: 156
content-type: application/json
date: Fri, 26 Jul 2019 18:39:40 GMT

{
    "error": {
        "code": -32602,
        "data": "Failed parsing args: invalid type: sequence, expected a string",
        "message": "Invalid params"
    },
    "id": "dontcare",
    "jsonrpc": "2.0"
}

```

ilblackdragon:
Reopening this, as it's true again.
Query, block, tx status and missing new RPCs like valiators, network_info and chunk

amgando:
this seems to have been resolved by #164

PR #164, opened 2020-01-22
document updates made by #1962

amgando:
partially addresses #48

### DIFF
--- a/docs/interaction/rpc.md
+++ b/docs/interaction/rpc.md
@@ -6,7 +6,14 @@ sidebar_label: RPC
 
 Standard JSON RPC 2.0 is used across the board.
 
-The following methods are available:
+Notes:
+
+- all methods are HTTP `POST`
+- endpoint URL varies by network
+  - for TestNet use `https://rpc.nearprotocol.com`
+  - for Staging use `https://staging-rpc.nearprotocol.com`
+
+You can see this interface defined in `nearcore` [here](https://github.com/nearprotocol/nearcore/blob/master/chain/jsonrpc/client/src/lib.rs#L185)
 
 ## Status
 
@@ -14,9 +21,7 @@ The following methods are available:
 
 Returns current status of the node
 
-- HTTP `POST`
-- URL `https://rpc.nearprotocol.com`
-- method `query`
+- method `query(path: string, data: string)`
 - params *none*
 
 ```bash
@@ -53,12 +58,10 @@ http post https://rpc.nearprotocol.com jsonrpc=2.0 method=status params:='[]' id
 
 ## Send transaction (async)
 
-`broadcast_tx_async`
+`broadcast_tx_async(signed_transaction: string)`
 
 Sends transaction and returns right away with the hash of the transaction in base58.
 
-- HTTP `POST`
-- URL `https://rpc.nearprotocol.com`
 - method `broadcast_tx_async`
 - params
   - (1) `<base 64 of the SignedTransaction>`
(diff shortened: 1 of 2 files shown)


docs/interaction/rpc.md
```diff
@@ -10,118 +10,508 @@ The following methods are available:
...
-`http post http://127.0.0.1:3030/ jsonrpc=2.0 method=broadcast_tx_async params:="[<base 64 of the SignedTransaction>]" id="dontcare"`
+- HTTP `POST`
+- URL `https://rpc.nearprotocol.com`
+- method `broadcast_tx_async`
+- params
+  - (1) `<base 64 of the SignedTransaction>`
+
+```bash
+http post https://rpc.nearprotocol.com jsonrpc=2.0 method=broadcast_tx_async params:='["<base 64 of the SignedTransaction>"]' id=dontcare
+```
```
ilblackdragon: let's add example output here as well.

docs/interaction/rpc.md
```diff
@@ -10,118 +10,508 @@ The following methods are available:
...
+      "validator_reward": "0"
+    }
+  },
+  "id": "dontcare"
+}
+```
+
+## Transaction Status Endpoint
+
+`tx(txHash: string)`
```
ilblackdragon: two arguments required, txHash + senderId

docs/interaction/rpc.md
```diff
@@ -10,118 +10,508 @@ The following methods are available:
...
+
+Queries status of a transaction by hash and returns `FinalTransactionResult`.
+
+- HTTP `POST`
+- URL `https://rpc.nearprotocol.com`
+- method `tx`
+- params
+  - (1) `<base 58 of transaction hash>` (see [NEAR Explorer](https://explorer.nearprotocol.com) for a valid transaction hash)
+  - (2) `<sender account id>` (ignored)
```
ilblackdragon: required -- is to determine which shard to look for transaction

docs/interaction/rpc.md
```diff
@@ -10,118 +10,508 @@ The following methods are available:
...
 
-`tx(hash: bytes)`: queries status of the transaction by hash, returns FinalTransactionResult that includes status, logs and result: `{"status": "Completed", "logs": [{"hash": "<hash>", "lines": [], "receipts": [], "result": null}]}`
+Queries for active validators on the network and returns details about them and the current state of validation on the blockchain
 
-`http post http://127.0.0.1:3030/ jsonrpc=2.0 method=tx params:=["<base 58 of transaction hash>", "<sender account id>"] id=dontcare`
+- HTTP `POST`
+- URL `https://rpc.nearprotocol.com`
+- method `validators`
+- params
+  - (1) `<base 58 block hash>` (see [status page](https://rpc.nearprotocol.com/status) for a valid `latest_block_hash`)
```
ilblackdragon: also can use "null" for latest

docs/interaction/rpc.md
```diff
@@ -10,118 +10,508 @@ The following methods are available:
...
+        "public_key": "ed25519:22skMptHjFWNyuEWY22ftn2AbLPSYpmYwGJRGwpNHbTV",
+        "shards": [
+          0
+        ],
+        "stake": "50445324683980935237733345"
+      }
+    ]
+  }
+}
+```
```
ilblackdragon: you can refer to the full list here - https://github.com/nearprotocol/nearcore/blob/master/chain/jsonrpc/client/src/lib.rs#L185

there are also:
 - chunk(bloch_height / block_hash)
 - gas_price('null' / block_height / block_hash)

docs/validator/staking.md
```diff
@@ -132,3 +132,18 @@ If you would like to see how much a validator is staking, you can run the comman
   storage_usage: 182
 }
 ```
+
+## Automatically re-staking
+
+This is implemented as a separate service that runs in parallel with `nearcore` and re-stakes if it sees that given validator might be kicked out or is already kicked out.
```
ilblackdragon: you can run this service either on the same machine or on a different machine for extra security to keep signing key separate.

docs/validator/staking.md
```diff
@@ -132,3 +132,18 @@ If you would like to see how much a validator is staking, you can run the comman
...
+
+Note that right now this uses `validator_key.json` to sign transactions as well.  We plan to add an option for a different signing key soon.
+
+```bash
+cargo build -p restaked --release
+./target/release/restaked \
+    --home=/path/to/validator_key.json \
```
ilblackdragon: let's use `--home=<path to folder with validator key>` because it's path to folder not to file.

docs/interaction/rpc.md
amgando: ✅

docs/interaction/rpc.md
amgando: ✅

docs/interaction/rpc.md
amgando: @ilblackdragon 

❓not sure how to get this to work

thrashing with a few formats didn't work: 
- `params:='["null"]'` -- "provided string contained invalid character 'l' at byte 2"
- `params:='[null]'` -- "Failed parsing args: invalid type: null, expected a string"
- `params:='null'` -- "Require at least one parameter"

note: this works just fine
```
http post https://rpc.nearprotocol.com jsonrpc=2.0 method=validators params:='["8ehA3NYL5uSF8zefbnqnz66twYJ45rfst6SrqBNv7oka"]' id=dontcare
```

docs/interaction/rpc.md
amgando: ✅

docs/interaction/rpc.md
amgando: ✅ on this one

also fixed incorrect `broadcast_tx_commit` output entry since that's what NEAR Shell uses when we call a function anyway.

docs/interaction/rpc.md
amgando: spoke w @frol and found this works fine on staging.  

`params:='[null]'` is correct

amgando: ✅

docs/validator/staking.md
amgando: ✅

docs/validator/staking.md
amgando: ✅

amgando:
closes #48

docs/interaction/rpc.md
```diff
@@ -6,122 +6,675 @@ sidebar_label: RPC
 
 Standard JSON RPC 2.0 is used across the board.
 
-The following methods are available:
+Notes:
+
+- all methods are HTTP `POST`
+- endpoint URL varies by network
+  - for TestNet use `https://rpc.nearprotocol.com`
+  - for Staging use `https://staging-rpc.nearprotocol.com`
```
janedegtiareva: Let's mention that this environment is not supposed to be stable. It may break at any moment. (do we even want an unstable environment in public docs?)

docs/interaction/rpc.md
```diff
@@ -6,122 +6,675 @@ sidebar_label: RPC
...
+### `contract/<account_id>`
+
+Returns full state of the contract (might be expensive if contract has large state).
+
 
-the `[...]` will  be replaced with a public key as an array.
+### `call/<account_id>/<method name>`
 
-Note, this is Tendermint-like compatibility, that should be refactored.
+Calls `<method name>` in contract `<account_id>` as view function with `data` as parameters.
```
frol: The second parameter is a JSON string encoded with base58 (NOT base64):

```
$ http post https://rpc.nearprotocol.com/ jsonrpc=2.0 id=dontcare method=query 'params:=["call/studio-heptm2q29/whoSaidHi","AQ4"]'
{
    "id": "dontcare",
    "jsonrpc": "2.0",
    "result": {
        "block_height": 1045855,
        "logs": [],
        "result": [34, 102, 114, 111, 108, 46, 110, 101, 97, 114, 34]  // This is a JSON string: "frol.near" (with quotes)
    }
}
```

`AQ4` is `{}` (empty JSON object) in base58 encoding. The output result is an array of bytes (numbers) of JSON.

docs/interaction/rpc.md
```diff
@@ -6,122 +6,675 @@ sidebar_label: RPC
...
+      }
     }
+  }
 }
 ```
 
-Where `value` is base58 encoded JSON of the account status: `'{"account_id":"test.near","nonce":0,"amount":1000000000000,"stake":50000000,"public_keys":[[...]],"code_hash":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="}'`
+### `contract/<account_id>`
+
+Returns full state of the contract (might be expensive if contract has large state).
```
frol: Heads up!

This method is currently in a broken state. @ilblackdragon has a PR to fix this https://github.com/nearprotocol/nearcore/pull/2075, and the API is:

```
$ http post http://localhost:3030 jsonrpc=2.0 method=query id=123 params:='["contract/my_token", ""]'
```
```js
{
    "id": "123",
    "jsonrpc": "2.0",
    "result": {
        "block_height": 2057,
        "values": [
            {
                "key": "U1RBVEU=",
                "value": "AgAAAGFpAQAAAAAAAAACAAAAYWsBAAAAAAAAAAIAAABhdqCGAQAAAAAAAAAAAAAAAAA="
            },
            {
                "key": "YWkMAAAAbXlfdmFsaWRhdG9y",
                "value": "AAAAAAAAAAA="
            },
            {
                "key": "YWsAAAAAAAAAAA==",
                "value": "DAAAAG15X3ZhbGlkYXRvcg=="
            },
            {
                "key": "YXYAAAAAAAAAAA==",
                "value": "oIYBAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
            }
        ]
    }
}
```

The second input parameter is base58-encoded state key prefix. It is worth noting that the new API #184 uses base64-encoded values. The output keys and values are encoded with base64 (the output is the same for current and the new APIs).

docs/interaction/rpc.md
amgando: i added a short message about instability.  the only reason i included it here is bc sometimes TestNet is down and the only way to get an RPC response (for testing, learning, etc) is via staging.

docs/interaction/rpc.md
amgando: ✅

docs/interaction/rpc.md
amgando: ✅

amgando:
rebased on master to include fixes for broken links and new link checker CI

PR #188, opened 2020-02-04
Create .travis.yml

vgrichina:


amgando:
this closes #189

amgando:
@vgrichina is there some easy way to ignore certain patterns of URL?

here's an example (there are a several) that is a valid use of a URL which is not available at time of validating the docs for CI


```
Getting links from: http://localhost:3000/docs/development/near-clitool
├───OK─── https://www.npmjs.com/get-npm
├───OK─── https://nodejs.org/en/download
├─BROKEN─ http://localhost:3030/ (ERRNO_ECONNREFUSED)
Finished! 95 links found. 92 excluded. 1 broken.
```

amgando:
i guess we can just add this and make sure all dev links match the pattern

```
npx broken-link-checker -gro -exclude "localhost:3030" http://localhost:3000
```

PR #191, opened 2020-02-05
fix broken links found using broken-link-checker

amgando:
preparing for #188


amgando:
@vgrichina note that this will report 1 broken link unless the CI command is edited as follows:

```
npx broken-link-checker -gro -exclude "localhost:3030" http://localhost:3000
```

Issue #158, opened 2020-01-15
[DOC] What is gas and its relation to NEAR tokens?


frol:
**Is your feature request related to a problem? Please describe.**
Gas vs NEAR tokens is a common question for newcomers.

**Describe the solution you'd like**
Here is my take on explaining this (we need to review this info as my understanding might be incomplete, and I think we should avoid comparison with "money"):

> gas is a unified unit of the computational cost, e.g. to store 1 byte of information is 1 gas and doing a computation of 2+2 takes 3 gas, etc. Token, on the other hand, is similar to "money", which you use to buy anything, including gas, which you spend to do computations.
>
> Thus, accounts have their balance in NEAR tokens and gas is bought on the fly when a transaction gets processed.
>
> So here is what happens. Account has its balance in NEAR tokens and when a user wants to send a transaction, he/she decides how much gas at most he/she is willing to spend for the transaction (some transactions require an almost constant amount of gas, but FunctionCalls are hard to predict); once the transaction reaches the network, it automatically "buys" gas for the current price, and executes the transaction; once the transaction is done, the leftover gas is converted back to NEAR tokens and the account gets only "charged" for the used gas.
>
> There is a prepaid gas for a transaction, and it only gets prepaid on the spot, so there is no way to prepay it for an account or something like that.

amgando:
the current mention of gas in our docs is incorrect and misleading on [this page](https://docs.nearprotocol.com/docs/roles/developer/examples/nearlib/introduction)

> **the cost of data storage and computation on the network is measured in units of gas which is a configurable bit of network economics but is usually vanishingly small**
> 
> It costs exactly 2 gas (at the time of this writing) to "compute the creation of a new account" on the network. Gas is measured in yoctoNEAR or yNEAR, which is 10^24 times smaller than NEAR. For a sense of scale, with just 2 of the 10 NEAR in your new account you could happily afford to create a new account for every grain of sand on the planet (actually estimates suggest about 10^21 grains of sand on earth so you could do this every day for 3 years, not just once) and still have some NEAR tokens left over for a rainy day. Gas is measured as very very very small fraction of NEAR tokens.

i'll update it to match your clearer explanation

frol:
I vote for having this fundamental information about what gas is in an “overview” section describing various terms used in Near instead of having this info buried in nearlib documentation.

/cc @nearmax

amgando:
how about in the same section as other "Key Concepts" like `Account` and `Transaction`?
[...]

PR #195, opened 2020-02-06
add content around key concept: gas

amgando:
content merged from 
- issue #158 
- [NEAR Whitepaper](https://nearprotocol.com/papers/the-official-near-white-paper/#economy) 
- various Discord posts

### DIFF
--- a/docs/concepts/gas.md
+++ b/docs/concepts/gas.md
@@ -1,0 +1,81 @@
+---
+id: gas
+title: Gas
+sidebar_label: Gas
+---
+
+Gas is a dynamic, unified unit of cost of computation and storage on the blockchain as well as network utilization among nodes.  
+
+Accounts automatically buy gas using NEAR tokens at various points in time as when they invoke a function (charged at the moment the transaction is processed) or pay storage rent (charged periodically based on amount of data stored by an account as well as a possible tax on having a short account name to prevent name squatting).
+
+This page covers these dynamics in more detail, starting with an an excerpt from the [NEAR Whitepaper](https://nearprotocol.com/papers/the-official-near-white-paper/) below:
+
+<blockquote class="info">
+<strong>NEAR Platform Economics -- What is Gas?</strong><br><br>
+
+Fundamentally, the NEAR platform is a marketplace between willing participants.
+
+On the supply side, operators of the validator nodes and other fundamental infrastructure need to be incentivized to provide these services which make up the community cloud.  On the demand side, the developers and end-users of the platform who are paying for its use need to be able to do so in a way which is simple, clear and consistent so it helps them.
+
+A blockchain-based cloud provides several specific resources to the applications which run atop it:
+
+- **Compute (CPU)**: This is the actual computer processing (and immediately available RAM) which run the code in a contract.
+- **Bandwidth ("Network")**: This is the network traffic between participants and users, including messages which submit transactions and those which propagate blocks.
+- **Storage**: Permanent data storage on the chain, typically expressed as a function of both storage space and time.   
+ 
+Existing blockchains like Ethereum account for all of these in a single up front transaction fee which represents a separate accounting for each of them but ultimately charges developers or users only once in a single fee.  
+
+**This is a high volatility fee commonly denominated in "gas".**
+
+Developers prefer predictable pricing so they can budget and provide prices for their end users.  The pricing for the above-mentioned resources on NEAR is an amount which is slowly adjusted based on system usage (and subject to the smoothing effect of resharding for extreme usage) rather than being fully auction-based. This means that a developer can more predictably know that the cost of running transactions or maintaining their storage.
+
+Initially, all of these resources will be priced and paid in terms of NEAR tokens. In the future, they may also be priced in terms of a stable currency denomination (for example a token pegged to the $USD).
+
+[read more here](https://nearprotocol.com/papers/the-official-near-white-paper/#economy)
+
+</blockquote>
+
+
+
+
+
+
+
+## How do I buy gas?
+
+You don't. 
+
+Accounts purchase gas automatically using NEAR tokens and at various points in time including:
+
+1. **when a transaction is processed** (ie. a function call is made, an account is created, tokens are transferred, etc) \
+    The amount of gas used by a transaction depends on the transaction itself and is fixed at genesis (ie. it requires 1 unit of gas to store 1 byte of information and it requires 3 units of gas to compute 2+2).  The actual *cost of this gas*, it's price in NEAR tokens is calculated dynamically and on an ongoing basis based on system load.
+
+2. **when an account is charged rent** to remain on the network  \
+    From the NEAR whitepaper, 
+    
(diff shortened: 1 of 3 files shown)


docs/concepts/gas.md
```diff
@@ -0,0 +1,200 @@
...
+</blockquote>
+
+## How do I buy gas?
+
+You don't. 
+
+Accounts purchase gas automatically using NEAR tokens and at various points in time including:
+
+1. **when a transaction is processed** (ie. a function call is made, an account is created, tokens are transferred, etc) \
+    The amount of gas used by a transaction depends on the transaction itself and is fixed at genesis (ie. it requires 1 unit of gas to store 1 byte of information and it requires 3 units of gas to compute 2+2).  The actual *cost of this gas*, it's price in NEAR tokens is calculated dynamically and on an ongoing basis based on system load.
```
frol: @nearmax Could you provide a more realistic example of how much storing / computing may cost?

@amgando I think the sentence "ie. it requires 1 unit of gas to store 1 byte" needs to be reworded to reflect that it only serves an example, not the actual exact numbers.

amgando:
@fckt, can we pair on this?

@nearmax made the following comment and suggested you can help clarify by looking at the implementation.

> _We do not purchase gas for the rent. We should be using the economic whitepaper for the guidance but the source of truth is the actual implementation._

docs/concepts/gas.md
nearmax: We shouldn't be giving the actual numbers, since we might not be event in the right order of magnitude.

docs/concepts/gas.md
```diff
@@ -0,0 +1,200 @@
+---
+id: gas
+title: Gas
+sidebar_label: Gas
+---
+
+Gas is a dynamic, unified unit of cost of computation and storage on the blockchain as well as network utilization among nodes.
```
fckt: Better to remove "storage" from the sentence, since in the implementation the sender buys gas to pay for computation. Storage rent price is in tokens, since it has to be static all the time.

docs/concepts/gas.md
```diff
@@ -0,0 +1,200 @@
+---
+id: gas
+title: Gas
+sidebar_label: Gas
+---
+
+Gas is a dynamic, unified unit of cost of computation and storage on the blockchain as well as network utilization among nodes.  
+
+Accounts automatically buy gas using NEAR tokens at various points in time as when they invoke a function (charged at the moment the transaction is processed) or pay storage rent (charged periodically based on amount of data stored by an account as well as a possible tax on having a short account name to prevent name squatting).
```
fckt: To be exact, the rent is not paid automatically.  It's accumulating and paid once account executes an incoming transaction or receipt. See:
https://github.com/nearprotocol/nearcore/blob/master/core/primitives/src/account.rs#L18
https://github.com/nearprotocol/nearcore/blob/master/runtime/runtime/src/lib.rs#L258
https://github.com/nearprotocol/nearcore/blob/master/runtime/runtime/src/lib.rs#L570

docs/concepts/gas.md
amgando: for sure, i can reword that as follows:

>  The amount of gas used by a transaction depends on the transaction itself and is fixed at genesis (ie. through some governance process, we may decide that it requires 1 unit of gas to store 1 byte of information but it requires 3 units of gas to compute the addition of two numbers, 3+2 -- note this is just a toy example, actual gas required for storage and compute can be verified experimentally using examples like those below).

but i'm still struggling to build my own confidence around the complete costs of all possible storage, compute and networking events in the system.  i think i recall someone was working on this, a kind of cost estimator, but i haven't seen anything about it.

i can see in my local `genesis.json` file that there are many settings for various bits about gas (`min_gas_price` and `max_gas_burnt` in general per block?, `runtime_config` has a dozen details, `wasm_config` has another dozen details, etc)

docs/concepts/gas.md
amgando: oh, nice.  thank you for this clarification

docs/concepts/gas.md
amgando: great, thank you

docs/concepts/gas.md
```diff
@@ -0,0 +1,200 @@
...
+Account has its balance in NEAR tokens
+
+A user decides to send a transaction to the network.  She decides how much gas at most she is willing to spend for the transaction and sets a gas limit for processing the transaction.  Some transactions require an almost constant amount of gas but function calls are hard to predict so a little extra never hurts since whatever remains unspent will be returned.
+
+The transaction reaches the network and, while being processed, it automatically "buys" enough gas to finish processing at the current gas prices.
+
+Once the transaction is processed, the related account is only "charged" for the gas that was used.  Any leftover gas is converted back to NEAR tokens on the account.
+
+
+## What about Prepaid Gas?
```
fckt: AFAIK the is no way to buy gas and accumulate it. As of onboarding, we use FunctionCall-limited access key which allows to spent a fixed amount of tokens for a specific method call

docs/concepts/gas.md
fckt: one small thing: saying "Gas is a dynamic" is confusing. Yes, gas amount per operation could change through the governance process, but in general - each operation has a fixed price in gas. Gas price is dynamic, but not gas itself.

amgando:
this PR will close #158

docs/concepts/gas.md
amgando: ok, took a closer look at the numbers in `genesis.json` and i see how any specific numbers can be wildly misleading.  i'll replace with this: 

> The amount of gas used by a transaction depends on the details of the transaction itself.  At genesis, the blockchain is configured with many different cost parameters (in units of gas).  Through some governance process, these parameters may be changed over time.  The actual *cost of this gas*, its price in NEAR tokens, is calculated dynamically and on an ongoing basis based on system load, inflation and other factors.

is this a better representation, @nearmax and @frol?

docs/concepts/gas.md
amgando: will rewrite to this.  is it better?

> Gas is a unified unit of cost of computation on the blockchain.  It is purchased at a variable price point (depending on system load) using NEAR tokens at the moment a transaction is processed.

docs/concepts/gas.md
amgando: will rewrite to this.  is it better?

> Accounts also pay storage rent which accumulates over time and is charged once the account submits a transaction for processing.

docs/concepts/gas.md
amgando: we have a few published statements about this idea of prepaid gas -- i was using this section to try to explain that what it actually means is that a developer funds an account that new users will draw from while onboarding.  once they switch to paid accounts, users would fund themselves.

docs/concepts/gas.md
bowenwang1996: Maybe worth mentioning that transactions other than function call will cost a fixed amount of gas (unless we change the protocol).

PR #201, opened 2020-02-09
Content.fix—cross contract comments test

mikedotexe:
Minor cleanup on an awesome article. I also removed a warning about the gas being exceeded because I see this issue has been addressed and closed here:
https://github.com/nearprotocol/nearlib/issues/176

### DIFF
--- a/docs/tutorials/how-to-write-contracts-that-talk-to-each-other.md
+++ b/docs/tutorials/how-to-write-contracts-that-talk-to-each-other.md
@@ -371,9 +371,9 @@ export function calculate({ a, b }): void {
 }
 ```
 
-This is a bit confusing because the contract is returning a promise, but because it's calling a function elsewhere, the compiler thinks that this is returning void.
+You may notice this function returns `void`, which is a bit confusing because the contract is returning a promise. This is because it's calling a function elsewhere and the compiler thinks it's void.
 
-(For now, you can set the return type of the function to `void`. In future releases this will be changed.)
+(In future releases this will be changed.)
 
 ## Step 6 - More Tests!
 
--- a/docs/tutorials/how-to-write-contracts-that-talk-to-each-other.md
+++ b/docs/tutorials/how-to-write-contracts-that-talk-to-each-other.md
@@ -94,19 +94,23 @@ We're interested in writing only one function for this example. A function that 
 import { context, storage } from "near-runtime-ts";
 
 export function addLongNumbers(a: string, b: string): string {
+  // Similar to long addition by hand, we start with the least significant digits first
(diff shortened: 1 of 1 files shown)


mikedotexe:
I've reverted that atomic commit removing the warning, and must confess I am quite confused now. I followed the directions in the cross-contract page using Studio. In the instructions, we start from the ready-made *Token Smart Contract.* Now I see that the `package.json` doesn't include `nearlib`. So I'm unsure how the `nearlib` issue with `new BN('…` plays into this.
This might be a question that's worth filing away in the back of our heads. There are some very minor changes in this PR that still add a teeny bit of value.

amgando:
oh i think you were right to remove that warning, it was my misunderstanding in the end and @vgrichina just suggested we wrap the gas with `BN` to make sure it's handled properly.

mikedotexe:
> oh i think you were right to remove that warning, it was my misunderstanding in the end and @vgrichina just suggested we wrap the gas with `BN` to make sure it's handled properly. i mean, doesn't it just work once you make this recommended edit? no gas limit warning at runtime?

I doesn't work when I add the `new BN(…` edit because it couldn't find `BN`. So then I looked at the imports for `nearlib` and didn't see it there. `BN` lives in `nearlib` as is discussed in the other issue. Then I went to the git blame for `NEARStudio`'s [Token Smart Contract dependencies](https://github.com/nearprotocol/NEARStudio/blob/master/templates/05-token-contract/package.json) to see if it used to have `nearlib` as a dependency at some point. (Thinking, perhaps, that it was updated.) Didn't see that either.

So I'm confused as to the how the mentioned issue (https://github.com/nearprotocol/nearlib/issues/176#issuecomment-577064933) has a link between `nearlib` and the smart contract in the tutorial, which is based on the Token Smart Contract as a base.
I feel like I just said a bunch of words that may weaved into tangled spaghetti. Happy to jump on a video chat to discuss if this is confusing.

mikedotexe:
Perhaps I can summarize with this picture. I don't get how `nearlib` plays into this issue, as specified in the linked issue.
Screenshot 2020-02-13 11 48 31

When I go to Studio and create the Token Smart Contract, where we start, not seeing nearlib. So perhaps we add that as a dependency and then use `new BN(`?
Screenshot 2020-02-13 11 51 33

mikedotexe:
@amgando did we decide to keep the gas warning in there and incorporate the rest of the comments?

amgando:
> _did we decide to keep the gas warning in there and incorporate the rest of the comments?_

@mikedotexe -- sure, let's merge this and revisit

PR #204, opened 2020-02-10
add dedicated page for create-near-app

amgando:
note, the image of folder structures for this page was generated using the [tree](https://rschu.me/list-a-directory-with-tree-command-on-mac-os-x-3b2d4c4a4827) command [this spreadsheet](https://docs.google.com/spreadsheets/d/17dVVj3hEBepXFPQprtizvBCczrt_FhJzzAv2djVd-gs/edit#gid=0)


amgando:
direct URL for content feedback: https://docs-pr-204.onrender.com/docs/quick-start/create-near-app

icerove:
good work, a little problem is that image is not clear (maybe the resolution problem), could it be text form instead of image? if not easy, more clear image better

amgando:
> good work, a little problem is that image is not clear (maybe the resolution problem), could it be text form instead of image? if not easy, more clear image better

i think an image makes more sense here.  will upload a higher resolution image

PR #209, opened 2020-02-11
remove or clarify all mention of TypeScript and replace with AssemblyScript

amgando:
Some unproductive confusion was reported by members of the team around the assumption that AssemblyScript was as feature rich as TypeScript.

This change aims to remove or explicitly qualify all mention on TypeScript throughout our documentation.

Where possible, TypeScript was renamed to AssemblyScript. When this was not possible, some qualification of the relationship between AssemblyScript and TypeScript was made:

ie. `AssemblyScript (a strict subset of TypeScript) ...`

Hopefully this eliminates or at least significantly reduces the chances for confusion between these two technologies.


willemneal:
> AssemblyScript is not a strict subset of TypeScript. It has features that TypeScript does not. Better call AssemblyScript a dialect of TypeScript.

I always described it as superset/subset, but I like dialect.

PR #212, opened 2020-02-12
correct mistake, receipt can hold one or more actions

amgando:
after speaking with @fckt about this during our offsite, it became clear that this diagram was incorrect.  it may still have errors but some improvement is always good


mikedotexe:
Travis is throwing errors that a couple YouTube links are broken, but they're not. This looks great. Will spend a little time seeing why this is, but going to merge this.

mikedotexe:
Not going to delete this remote branch just yet in case it's helpful in tracking down Travis issue.

PR #228, opened 2020-02-18
reduce code-boxes

lazaridiscom:
1st pass, can be further simplified


amgando:
nice, thank you.  i imagine this makes the page easier on the eyes too.  please keep the contributions coming @lazaridiscom

PR #229, opened 2020-02-18
add links, remove repo list

lazaridiscom:
- add link to repo overview of near github organization
- remove repo list (in favor or link to overview)
- add link to all "goog first issues"

### DIFF
--- a/docs/contribution/technical-contribution.md
+++ b/docs/contribution/technical-contribution.md
@@ -16,23 +16,7 @@ We'll always need help improving documentation, creating tutorials, and writing 
 
 ## Open Issues in NEAR repositories
 
-There are various repo's you can contribute to. If you're looking to collaborate and want to find easy tasks to start, look at the issues we marked as "Good first issue".
-
-[nearcore](https://github.com/nearprotocol/nearcore/issues) - Official NEAR Protocol client
-
-[near-shell](https://github.com/nearprotocol/near-shell/issues) - General purpose command line tools for interacting with NEAR Protocol
-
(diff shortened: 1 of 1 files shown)


amgando:
thanks for this @lazaridiscom!  appreciate the call out to good first issues and the top level github repo

regarding
> _remove repo list_

i think it's useful to have this list here, isn't it?  otherwise newcomers are looking at over 100 repos and scanning the top of the list for most recent activity to get a sense of the possible scope of work.

lazaridiscom:
>remove repo list (in favor or link to overview)

https://github.com/nearprotocol

Contains pinned repos and repo list => all a visitor needs

no need to duplicate info and increase tho doc size unnecessarily.

amgando:
> _no need to duplicate info and increase tho doc size unnecessarily._

while i don't think size of the document is important in this case, duplicating the list of key repos from pinned repos does seem like maintenance overhead, true.

but GitHub only lets us pin 6 repos to the top of the org landing page and we have more than 6 repos that we want to introduce to potential contributors.

let's try merging your changes for now and collect feedback.

thanks for the edits!

lazaridiscom:
> while i don't think size of the document is important in this case,

Keep docs compact, especially if they target professional folks. No need to exhaust them with redundant (possibly patronizing) talk.

If you want to address more casual contributors, possibly ones who try to contribute for the first time, write an additional intro and link to it. "For beginner contributors". The risk here is that later on, the team has to deal with beginner-coders (e.g. a blockchain project is not the best code-base to start a contributor career).

PR #233, opened 2020-02-19
document all possible errors to close #170

### LINKED ISSUE #170: When transaction is going to be bounced
We need an exhaustive documentation that explains all the stages when transaction can be rejected, under what conditions it can be rejected, and what error message we are going to observe through RPC when this happens. This should take into account the storage and the account name rent. Note, it is crucial that this documentation is exhaustive, because our partners will heavily rely on its exhaustiveness.

Transaction currently can be rejected in multiple places:
* Before it is included in the mempool;
* After it is pulled from the mempool;
* When chunk is executed.

@evgenykuzyakov do you remember any other places where we bounce transactions?

amgando:


amgando:
@nearmax please have a look to see if this satisfies #170

PR #238, opened 2020-02-21
add code of conduct, contributing, license

mikedotexe:
Addresses: https://github.com/nearprotocol/docs/issues/220
This is a request that came in a while ago, getting to it with the repos in closest reach.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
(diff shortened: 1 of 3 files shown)


mikedotexe:
> @mikedotexe -- the only thing that i would be concerned about is whether these standards are drawn from the same sources that we're using around the org. especially license. if this is what we're using on all repos, then good to go. if not, then that would be a blocker for me.

Yes, I took those files from `nearcore` as requested by Illia on Slack here:
https://near-protocol.slack.com/archives/CCCL1RK18/p1580932434049600?thread_ts=1580932434.049600
https://github.com/nearprotocol/nearcore/tree/stable

PR #244, opened 2020-02-27
exclude youtube from link-checker as it throws 429's

mikedotexe:
also a redirect to youtube: near.ai/wbs

---
Travis has been throwing many 429 errors because of too many requests when it tries to check broken links. This excludes youtube links and the link mentioned in the first line of this PR.


mikedotexe:
Sweet, looks like it's good to go. Thanks for the discussion and lead on this, Sherif

PR #245, opened 2020-02-27
remove instruction to checkout staging on local setup

mikedotexe:
Very simple, as the title says. There must have once been a time where it made sense to checkout staging, but I've been instructed to remove this.

### DIFF
--- a/docs/local-setup/running-testnet.md
+++ b/docs/local-setup/running-testnet.md
@@ -44,11 +44,6 @@ git clone https://github.com/nearprotocol/nearcore.git
 cd nearcore
 ```
 
-`checkout` to latest development version of start script:
-```bash
-git checkout staging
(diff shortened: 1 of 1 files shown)


damons:
Since we've changed the branching and code promotion scheme for nearcore, we need to make sure we instruct users to use our most stable network and code by default.  This will be the `stable` branch.

mikedotexe:
> Since we've changed the branching and code promotion scheme for nearcore, we need to make sure we instruct users to use our most stable network and code by default. This will be the `stable` branch.

goood catch. changed

mikedotexe:
just had a user get confused [about this on Discord](https://discordapp.com/channels/490367152054992913/490367152054992917/682799478507569161)
Going to merge this

PR #250, opened 2020-03-03
add stackoverflow nudge

amgando:


website/pages/en/index.js
```diff
@@ -201,6 +206,18 @@ class HomeSplash extends React.Component {
...
+          <MarkdownBlock>### Have a Question about Anything NEAR? </MarkdownBlock><br/>
+          <MarkdownBlock>**(1) Is your question already posted on StackOverflow?**</MarkdownBlock><br/>
+          <MarkdownBlock>We pay very close attention to [questions tagged with `nearprotocol` on StackOverflow](https://stackoverflow.com/tags/nearprotocol).</MarkdownBlock><br/>
+          <MarkdownBlock>**(2) Do you need help posting your question on StackOverflow?**</MarkdownBlock><br/>
+          <MarkdownBlock>We understand that sometimes we don't know enough to ask a meaningful question -- what we need is a conversation to discover the right questions. In this case just find us on [Telegram](https://t.me/cryptonear) or [Discord](http://near.chat). We'll be happy to help and appreciate your support in reposting our discussion to StackOverflow afterwards.</MarkdownBlock><br/>
+          <MarkdownBlock>**(3) Should we be doing something differently?**</MarkdownBlock><br/>
+          <MarkdownBlock>If you have any recommendations for us on how to improve our community education and support channels, we're wide open to your feedback.</MarkdownBlock>
```
chadoh: You can do this with one `MarkdownBlock` 😄 

<MarkdownBlock>
  ### Have a Question about Anything NEAR? 

  **(1) Is your question already posted on StackOverflow?**

  ...
</MarkdownBlock>

website/pages/en/index.js
```diff
@@ -201,6 +206,18 @@ class HomeSplash extends React.Component {
           </div>
           </div>
       </SplashContainer>
+      <div align="center">
+        <div align="left" style={SOreminder}>
+          <MarkdownBlock>### Have a Question about Anything NEAR? </MarkdownBlock><br/>
+          <MarkdownBlock>**(1) Is your question already posted on StackOverflow?**</MarkdownBlock><br/>
+          <MarkdownBlock>We pay very close attention to [questions tagged with `nearprotocol` on StackOverflow](https://stackoverflow.com/tags/nearprotocol).</MarkdownBlock><br/>
+          <MarkdownBlock>**(2) Do you need help posting your question on StackOverflow?**</MarkdownBlock><br/>
+          <MarkdownBlock>We understand that sometimes we don't know enough to ask a meaningful question -- what we need is a conversation to discover the right questions. In this case just find us on [Telegram](https://t.me/cryptonear) or [Discord](http://near.chat). We'll be happy to help and appreciate your support in reposting our discussion to StackOverflow afterwards.</MarkdownBlock><br/>
```
chadoh: Let's use a real dash of some sort (en or em) rather than the `--`

website/pages/en/index.js
```diff
@@ -201,6 +206,18 @@ class HomeSplash extends React.Component {
           </div>
           </div>
       </SplashContainer>
+      <div align="center">
+        <div align="left" style={SOreminder}>
```
chadoh: The style looks uninspired – should we get design input on this before we post it?

amgando:
@chadoh -- thanks so much for the comments.  

i agree, there's more work to do here if we don't want it to look bad.  

re: `<MarkdownBlock>` repetition -- can you verify this locally?  a single block on my machine wasn't formatting things the way i would expect them to.  although i tried a few variations of a single markdownblock locally, i may have missed something?

website/pages/en/index.js
chadoh: > re: <MarkdownBlock> repetition -- can you verify this locally? a single block on my machine wasn't formatting things the way i would expect them to. although i tried a few variations of a single markdownblock locally, i may have missed something?

@amgando I looked into this a bit and couldn't figure out how to get it to work, but I've pushed a commit that removes `MarkdownBlock` in favor of straight html elements (it also removes the other occurrence of `MarkdownBlock` that had been in this file, since it's no longer used). Let me know what you think.

website/pages/en/index.js
chadoh: The `width: 90%` is my least favorite part of these styles – the width of this SO disclaimer doesn't match the width of the content above, or anything else on the page. However, I noticed that the mobile styles on this page could use some deeper cleanup, and that makes me wonder if the `width: 90%` is fine for now.

PR #252, opened 2020-03-04
add RPC named params and finality flag. close #184

### LINKED ISSUE #184: [DOC] Document extended/new query RPC API
There is an upcoming improvement in `query` RPC API, which allows querying account info, access key(s), and state by a given block ID (block height or block hash).

See more info and examples here: https://github.com/nearprotocol/nearcore/pull/2038#issuecomment-580444603

amgando:


amgando:
@frol see the page here: https://docs-pr-252.onrender.com/docs/api/rpc-params

docs/api/rpc-params.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Using block height
+
+```bash
+http post https://staging-rpc.nearprotocol.com jsonrpc=2.0 id=dontcare \
+    method=query \
+    'params:={
+                "request_type": "view_access_key_list", 
+                "account_id": "node0", 
+                "block_id": 33442,
```
frol: `block_id` and `finality` are mutually exclusive parameters (I guess the initial implementation was misleading, we have already addressed this in the source code). Thus, if you intend to query for a block height, you only specify `"block_id": 33442`, and don't include `"finality"` at all.

docs/api/rpc-params.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+### Using block hash
+```bash
+http post https://staging-rpc.nearprotocol.com jsonrpc=2.0 id=dontcare \
+    method=query \
+    'params:={
+                "request_type": "view_access_key_list", 
+                "account_id": "node0", 
+                "block_id": 33442,
```
frol: It should be a hash here.

P.S. remove `finality` (^^^)

docs/api/rpc-params.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Using `optimistic` finality
+
+```bash
+http post https://staging-rpc.nearprotocol.com jsonrpc=2.0 id=dontcare \
+    method=query \
+    'params:={
+                "request_type": "view_access_key_list",
+                "account_id": "node0", 
+                "block_id": 33442,
```
frol: Remove `block_id` (^^^)

docs/api/rpc-params.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Using `near-final` finality
+
+```bash
+http post https://staging-rpc.nearprotocol.com jsonrpc=2.0 id=dontcare \
+    method=query \
+    'params:={
+                "request_type": "view_access_key_list",
+                "account_id": "node0", 
+                "block_id": 33442,
```
frol: Remove `block_id` (^^^)

docs/api/rpc-params.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Using `final` finality
+
+```bash
+http post https://staging-rpc.nearprotocol.com jsonrpc=2.0 id=dontcare \
+    method=query \
+    'params:={
+                "request_type": "view_access_key_list",
+                "account_id": "node0", 
+                "block_id": 33442,
```
frol: Remove `block_id` (^^^)

docs/api/rpc-params.md
```diff
@@ -0,0 +1,158 @@
...
+                "finality": "final"
+             }'
+```
+
+<blockquote class="info">
+<strong>did you know?</strong><br><br>
```
frol: Since it is the only place we document this, I guess it is quite sarcastic to expect people to know this :smile:

I think we want to document this in the interaction/rpc document, and then this "did you know?" section will make sense, and it can be short with just a few words and a link to the section in the interaction/rpc document.

docs/api/rpc-params.md
```diff
@@ -0,0 +1,158 @@
...
+
+The `request_type` parameter can take any one of the following values (and associated parameters)
+
+```rust
+view_account          account_id: string
+view_state            account_id: string, prefix: string /* base64 */
+view_access_key       account_id: string, public_key: string
+view_access_key_list  account_id: string
+call_function         account_id: string, method_name: string, args: string /* base64 */
+```
```
frol: It is hard to find the right way to document this, but I think this block can be hard to read since it is not immediately clear that this is a table. Maybe a nested list or a list of tables will serve us better.

Here is an example of a table I have in mind:

image

Here is the nested list: https://github.com/nearprotocol/nearcore/pull/2148 (under API section)

Neither of the options seems ideal given that we have 1-3 arguments here. Maybe, we should just be sparser and go with a full example of all the parameters for each of the available types.

docs/api/rpc-params.md
```diff
@@ -0,0 +1,158 @@
...
+
+For a concrete example we can use the NEAR StageNet environment.
+
+1. The first two examples above are based on choosing a `request_type` of `view_access_key_list` which in turn requires `account_id` as a parameter.  
+
+2. As of [nearcore/2127](https://github.com/nearprotocol/nearcore/pull/2127) we also need to specify finality as one of `optimistic`, `near-final` or `final`.
```
frol: Given the above clarification, we seem need to refrase this comment to indicate that one **can** specify faniality instead of an exact block ID.

docs/api/rpc-params.md
```diff
@@ -0,0 +1,158 @@
...
+3. We can fetch the latest block hash from https://staging-rpc.nearprotocol.com/status
+
+```bash
+http post https://staging-rpc.nearprotocol.com jsonrpc=2.0 id=dontcare \
+    method=query \
+    'params:={
+                "request_type": "view_access_key_list",
+                "account_id": "node0", 
+                "block_id": "Ub5ctYJgeka2LGRj3h9VCTYA4BWg73bpMN8qvfRzd3R",
+                "finality": "optimistic"
```
frol: Choose one :smile:

website/sidebars.json
```diff
@@ -145,7 +145,8 @@
       "api/quickstart"
     ],
     "JSON RPC": [
-      "interaction/rpc"
+      "interaction/rpc",
+      "api/rpc-params"
```
frol: I don't feel good about having an endless document of all the APIs, but having such a split does not make a lot of sense to me unless we choose to only cover the fact that we use block ID || finality, and migrate to an object-based API (instead of position-based).

docs/api/rpc-params.md
amgando: ✅

docs/api/rpc-params.md
amgando: ✅

docs/api/rpc-params.md
amgando: ✅

docs/api/rpc-params.md
amgando: ✅

docs/api/rpc-params.md
amgando: ✅

docs/api/rpc-params.md
amgando: i think this is an English thing -- in the US, it's quite natural to say to someone, "hey, did you know [random fact X about Y]?" and there's no sarcasm at all.  it's like news or an update or something.  here in our docs this is a consistent blue callout in many places that brings attention to something that would be useful to know

docs/api/rpc-params.md
amgando: ✅

docs/api/rpc-params.md
amgando: ✅

docs/api/rpc-params.md
amgando: ✅

website/sidebars.json
amgando: i was separating this as part of a preview (what's on stagenet, for example, vs testnet)

docs/api/rpc-params.md
frol: @amgando TIL. Thanks!

I think there is still a subtle difference in the case when this is *the only* place you could learn about the fact. Though, I trust your judgement.

PR #253, opened 2020-03-04
add rpc experimental methods. close #184

### LINKED ISSUE #184: [DOC] Document extended/new query RPC API

amgando:


amgando:
@frol see here: https://docs-pr-253.onrender.com/docs/api/rpc-experimental

docs/api/rpc-experimental.md
```diff
@@ -0,0 +1,305 @@
...
+
+This page refers to a **technical preview**.
+
+These features are not currently available on TestNet. They will be available in an upcoming release.
+
+</blockquote>
+
+Two new methods will be exposed:
+
+### `EXPERIMENTAL_genesis_config`
```
frol: We should align the style of the headers with the "RPC" document, where we use human-readable titles and pseudo-calls like `EXPERIMENTAL_genesis_config()`, `EXPERIMENTAL_genesis_records(pagination: Pagination)`

amgando: ✅

PR #262, opened 2020-03-09
clarify prepaid gas

amgando:
based on this discussion in Telegram: https://t.me/cryptonear/9623


amgando:
live preview: https://docs-pr-262.onrender.com/docs/concepts/gas#what-about-prepaid-gas

image

PR #264, opened 2020-03-13
Content.fix  validator economics

jimmy3dita:
Needs review by @bowenwang1996 for technicals, and help by @amgando with my spaghetti-pizza-mandolino English.

Also, I didn't test 100% how the website will render the sidebar label and the links, so I'll keep an eye to this PR to quickly iterate.


docs/validator/economics.md
```diff
@@ -0,0 +1,134 @@
+---
+id: economics
+title: The Economics of a Validator
+sidebar_label: Understanding Economics
+---
+## Overview
+
+NEAR is a sharded, developer-friendly, proof-of-stake blockchain. It will use a technique called 'dynamic sharding' that splits the network, so that much of the computation is done in parallel. With the increase of users and their transactions, NEAR will dynamically introduce additional shards.
```
bowenwang1996: I think it's a bit strange to highlight something that we haven't implemented yet in the very beginning.

docs/validator/economics.md
```diff
@@ -0,0 +1,134 @@
...
+
+As a validator, you need to secure three classes of resources:
+1. NEAR tokens to stake
+2. Nodes infrastructure
+3. Continuous operation
+
+
+## 1. NEAR tokens to stake
+
+NEAR protocol determines validators at the end of each `epoch`, ordering them by the size of their stake. Current validators are re-elected automatically, by *re-staking* their tokens plus the accrued rewards; new validators may replace an existing one if they offer a more significant stake.
```
bowenwang1996: Not sure what you mean by "ordering them by the size of their stake". The order of validators is not determined by their stake.
Also it is imprecise to say that current validator will be re-elected automatically. It is better to say that they will automatically be enrolled in the auction that determines the validators for the future epochs.

docs/validator/economics.md
```diff
@@ -0,0 +1,134 @@
...
+    A large enough stake can obtain more than one seat, multiplying the rewards. 
+</blockquote>
+
+### Stake your NEAR tokens
+
+Use [NEAR Shell](../development/near-clitool.md) to issue a staking transaction and append your transaction to the `proposals`. At the end of the `epoch`, if your proposal is sufficient, your node will become a validator - and will begin to generate chunks/blocks after one more epoch. You will be able to monitor the `proposals`, and eventually top-up your transaction to increase your odds.
+
+### Stake delegated NEAR tokens
+
+NEAR leverages *smart contracts* for delegation. Therefore, every validator should implement its staking contract, or use one already available among the community.
```
bowenwang1996: Not every validator needs to do this. Only those who wish to offer delegation or delegate their stake.

docs/validator/economics.md
```diff
@@ -0,0 +1,134 @@
...
+</blockquote>
+
+
+## 2. Nodes infrastructure
+
+Once you obtained your NEAR stake, you need an infrastructure to serve the network securely.
+At the beginning of a new `epoch`, the protocol randomly assigns every validator to one of these roles:
+- Chunk Producer, for collecting transactions and applying them to the state
+- Block Producers, for generating the single blocks containing the current chunks 
+- Fishermen (aka 'hidden' validators), for challenging the state to verify the work of other validators
```
bowenwang1996: This is not true. Chunk producers are always block producers as well. Fishermen are rather separate. They are people who stake voluntarily for the safety of the network. They do not gain any reward through staking.

docs/validator/economics.md
```diff
@@ -0,0 +1,134 @@
...
+
+Whether the case, every validator must provide **availability** and **consistency**.
+
+### Availability
+
+The protocol requires a `MINIMUM_THRESHOLD` of validating signatures, proportional to the median chunk/block producer in the `epoch` and the number of your *seats*. If your node falls below `90 %`, it will be *kicked out* at the end of the epoch and you will have to issue the staking `proposal` again.
+In practical terms, you have to:
+Wait for the end of your failed epoch to unlock your stake
```
bowenwang1996: Doesn't have to wait until the end. As soon as you realize that you wouldn't be able to meet the threshold in the current epoch you can issue a new staking transaction.

docs/validator/economics.md
jimmy3dita: Happy to review the language to be more specific. I thought that winning proposals are put into an ordered array that we called `validatorAssignments`. If not by the stake, how are they ordered? Aren't the validators elected by the size of their stake - such that higher stakes come first?

docs/validator/economics.md
jimmy3dita: Awesome, so can I simply call them block producers and avoid any confusion around the term "chunks". Do you think that leaving a small footnote to clarify that will work? It may link to the part of the Nightshade Paper that instead separates these two.

Regarding fishermen: are they anyway submitting their proposal, or they simply run a node in a different state, by issuing a particular staking transaction?

docs/validator/economics.md
bowenwang1996: They are elected by the size of their stake, but the order of elected validators is not determined by their stake. We use the randomness seed from the randomness beacon to shuffle validators.

docs/validator/economics.md
bowenwang1996: > Do you think that leaving a small footnote to clarify that will work?

I think that's fine.


> Regarding fishermen: are they anyway submitting their proposal, or they simply run a node in a different state, by issuing a particular staking transaction?

They also submit proposal but there is a threshold for becoming fishermen which will be much smaller than the threshold for becoming validators. So for people who want to become fishermen, it makes sense for them to stake exactly the threshold amount.

docs/validator/economics.md
```diff
@@ -46,22 +46,19 @@ This option is still work in progress, so please follow the [NEP on GitHub](http
...
-Whether the case, every validator must provide **availability** and **consistency**.
+Once you obtained your NEAR stake, you need to setup the infrastructure to serve NEAR network.
+At a high level, your nodes must provide:
+- Availability
+- Consistency
 
 ### Availability
 
-The protocol requires a `MINIMUM_THRESHOLD` of validating signatures, proportional to the median chunk/block producer in the `epoch` and the number of your *seats*. If your node falls below `90 %`, it will be *kicked out* at the end of the epoch and you will have to issue the staking `proposal` again.
+The protocol requires your nodes to provide a `MINIMUM_THRESHOLD` of signatures, proportional to the median block producer in the `epoch`, and the number of your *seats*. If your node falls below `90 %`, it will be *kicked out* at the end of the epoch, and you will have to issue a staking `proposal` again.
```
bowenwang1996: This is somewhat misleading. It is better to say that the validators are required to produce a certain number of blocks and will be kicked out otherwise.

docs/validator/economics.md
```diff
@@ -46,22 +46,19 @@ This option is still work in progress, so please follow the [NEP on GitHub](http
...
 In practical terms, you have to:
-Wait for the end of your failed epoch to unlock your stake
-Stake your `proposal` again, and wait for the conclusion of the second epoch
-If your `proposal` succeed, wait for the third epoch, which is necessary to sync your node with the assigned shard
-With 730 epochs per year, you will miss at least 24 hours of rewards - therefore plan to have node redundancy over several locations and eventual connectivity fallback.
+- Re-stake your `proposal` again before the end of the failed epoch, and
+- If your `proposal` succeed, wait for the next epoch, which is necessary to sync your node with the assigned shard
```
bowenwang1996: "wait for the next epoch" can be ambiguous. I think it is better to say "wait for one more epoch".

docs/validator/economics.md
```diff
@@ -5,7 +5,7 @@ sidebar_label: Understanding Economics
 ---
 ## Overview
 
-NEAR is a sharded, developer-friendly, proof-of-stake blockchain. It will use a technique called 'dynamic sharding' that splits the network, so that much of the computation is done in parallel. With the increase of users and their transactions, NEAR will dynamically introduce additional shards.
+NEAR is a sharded, developer-friendly, proof-of-stake blockchain. Once fully deployed, its nodes will use a technique called 'dynamic sharding' that splits the network, so that much of the computation is done in parallel. With the increase of users and their transactions, NEAR will dynamically introduce additional shards.
```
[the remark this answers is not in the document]
jimmy3dita: Further reviewed

docs/validator/economics.md
```diff
@@ -17,24 +17,24 @@ As a validator, you need to secure three classes of resources:
 
 ## 1. NEAR tokens to stake
 
-NEAR protocol determines validators at the end of each `epoch`, ordering them by the size of their stake. Current validators are re-elected automatically, by *re-staking* their tokens plus the accrued rewards; new validators may replace an existing one if they offer a more significant stake.
+NEAR protocol determines validators before every new `epoch`, electing them by the size of their stake. Current validators are enrolled automatically and the end of the `epoch`, by *re-staking* their tokens plus the accrued rewards; new validators may replace an existing one if they offer a higher stake.
 
-As a validator, you can obtain a *seat* by:
+You can obtain a validator *seat* by:
```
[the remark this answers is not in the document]
jimmy3dita: Agree, it may be confusing. Reworded

docs/validator/economics.md
```diff
@@ -30,11 +30,11 @@ You can obtain a validator *seat* by:
 
 ### Stake your NEAR tokens
 
-Use [NEAR Shell](../development/near-clitool.md) to append your staking transaction to the current `proposals`. At the end of the `epoch`, if your proposal is sufficient, your node will become a validator - and will begin to generate rewards after one more epoch. If needed, you can increase anytime your `proposal` by signing a larger staking transaction using the same *stakingKey*.
+Use [NEAR Shell](../development/near-clitool.md) to generate your staking transaction and issue your `proposal` to become validator. At the end of the `epoch`, if your proposal is large enough, your node will become a validator - and will begin to generate rewards after one more epoch. If needed, you can increase anytime your `proposal` by signing a larger staking transaction using the same *stakingKey*.
```
bowenwang1996: I suggest you switch "generate" and "issue" in this sentence. It makes more sense that way.

docs/validator/economics.md
```diff
@@ -30,11 +30,11 @@ You can obtain a validator *seat* by:
...
 
-### Stake delegated NEAR tokens
+### Stake third-party NEAR tokens via delegation
```
bowenwang1996: Should we mention here that delegation is custodial?

jimmy3dita: Is this a trap? 😆
Delegation via smart contracts shouldn't need to be custodial - a delegator loads his tokens into it and begins to compound rewards (minus fees for the validator). It may become custodial only if the validator must know delegators' private keys.
From what I can see, it can be custodial only if the wallet is hosted within "some service" that won't give you the private keys of your wallet - and it's also being a validator.

docs/validator/economics.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Stake third-party NEAR tokens via delegation
+
+NEAR leverages *smart contracts* to enable the staking of third-party tokens through delegation. Therefore, every validator may implement its staking contract, or use one made available by the community. 
+This option is still work in progress, so please follow the [NEP on GitHub](https://github.com/nearprotocol/NEPs/blob/staking-contract/text/0000-staking-contract.md) for the specifications, and add your comments in the 'Riskless Delegation' discussion on [NEAR's Commonwealth](https://commonwealth.im/near/proposal/discussion/357-riskless-delegation-aka-tezos-delegation).
+
+<blockquote class="warning">
+    <strong>heads up</strong><br><br>
+    In the beginning, the network will run on a limited number of shards, with all validators competing for a limited number of seats.
+    As a simple rule-of-thumb, distributing 40% of the total supply (400 million out of 1 billion NEAR tokens) between 100 seats will need an average of 4 million NEAR tokens per validator.
```
ilblackdragon: will need at least 4m $NEAR to become a validator and take at least one seat.

docs/validator/economics.md
```diff
@@ -0,0 +1,132 @@
...
+A high median will have fewer validators with large stakes, thus introducing higher barriers of entry for any new `proposal`. On the other hand, a low median decreases the needed stake, introducing more turnaround and competition between validators.
+You may expect the latter, assuming that delegators will spread their stake across multiple validators to hedge risks.
+
+**Protocol fees and inflation**
+In detail, rewards follow the formula `total_supply * 0.05/730` of NEAR tokens at the end of every `epoch` (producing 5% of inflation per year). However, the protocol *burns* the costs of `tx_fees + state_rent` at the end of each block - such that rewards increase `total_supply` and fees reduce it. 
+Specifically, the protocol mints new tokens at the end of each epoch, while burning `tx_fees + state_rent` at each block.
+
+<blockquote class="info">
+    <strong>did you know?</strong><br><br>
+    Beyond a certain level of usage, the protocol will burn more tokens than minting new ones, decreasing NEAR's total supply.
```
ilblackdragon: I think generally, it's better to say that inflation rate depends on the amount of usage network gets.
And it's possible that total supply starts to decrease if the amount of fees collect overcomes the 5% annual security cost paid to validators.

docs/validator/economics.md
```diff
@@ -0,0 +1,132 @@
...
+    <strong>did you know?</strong><br><br>
+    The protocol automatically re-stakes your rewards at the end of each epoch, so you may expect up to 5% of additional stake every year.
+</blockquote>
+
+**Median of the stake across validators**
+A high median will have fewer validators with large stakes, thus introducing higher barriers of entry for any new `proposal`. On the other hand, a low median decreases the needed stake, introducing more turnaround and competition between validators.
+You may expect the latter, assuming that delegators will spread their stake across multiple validators to hedge risks.
+
+**Protocol fees and inflation**
+In detail, rewards follow the formula `total_supply * 0.05/730` of NEAR tokens at the end of every `epoch` (producing 5% of inflation per year). However, the protocol *burns* the costs of `tx_fees + state_rent` at the end of each block - such that rewards increase `total_supply` and fees reduce it.
```
ilblackdragon: I would not use 5% of inflation per year, because later you say that inflation is lower. May be say "producing 5% of new tokens per year".
Also there is a bit of computation weirdness here, because we issue 0.00009433% of total supply every epoch, and as total supply fluctuate, it's not actually going to be for example 5% of total supply on day 1.

docs/validator/economics.md
```diff
@@ -0,0 +1,132 @@
...
+    <strong>did you know?</strong><br><br>
+    Beyond a certain level of usage, the protocol will burn more tokens than minting new ones, decreasing NEAR's total supply.
+</blockquote>
+
+**Exchange value of NEAR tokens**
+Validators pay operating expenses in their local currency, and they may have to exchange their NEAR rewards. Most Proof-of-Work miners face similar problems (e.g., energy bills and maintenance personnel), so we can apply the same principles: an appreciating token leaves more operating margin opening opportunities for new validators. On the other hand, a depreciating token may 'squeeze out' the least efficient validators, and centralize the stake.
+Since rewards are fixed, this can be considered a *zero-sum game* influenced by the exchange value of the token.
+
+**Opportunity cost of other PoS tokens**
+Assuming zero barriers to acquire or exchange PoS tokens, every delegator will measure inflation, price fluctuations and potential returns of the same stake across multiple Proof-of-Stake ledgers. Therefore, you have to consider a return threshold, which may impose fee adjustments to stay in the market.
```
ilblackdragon: Hm, I'm not sure this is worth pointing out here.
The assumption is totally wrong, as buying even a $1m of tokens is extremely hard in anything that is not BTC.
Plus if you have delegation, users who delegate may have other motives (like using the network) on top.
My point, I don't think we should have this in our docs

docs/validator/economics.md
```diff
@@ -0,0 +1,132 @@
...
+**Protocol fees and inflation**
+In detail, rewards follow the formula `total_supply * 0.05/730` of NEAR tokens at the end of every `epoch` (producing 5% of inflation per year). However, the protocol *burns* the costs of `tx_fees + state_rent` at the end of each block - such that rewards increase `total_supply` and fees reduce it. 
+Specifically, the protocol mints new tokens at the end of each epoch, while burning `tx_fees + state_rent` at each block.
+
+<blockquote class="info">
+    <strong>did you know?</strong><br><br>
+    Beyond a certain level of usage, the protocol will burn more tokens than minting new ones, decreasing NEAR's total supply.
+</blockquote>
+
+**Exchange value of NEAR tokens**
```
ilblackdragon: also not sure if this belongs in our docs.

docs/validator/economics.md
```diff
@@ -0,0 +1,132 @@
...
+Since rewards are fixed, this can be considered a *zero-sum game* influenced by the exchange value of the token.
+
+**Opportunity cost of other PoS tokens**
+Assuming zero barriers to acquire or exchange PoS tokens, every delegator will measure inflation, price fluctuations and potential returns of the same stake across multiple Proof-of-Stake ledgers. Therefore, you have to consider a return threshold, which may impose fee adjustments to stay in the market.
+
+## Conclusions
+
+In the beginning, financial commitment and complexity of operations will privilege professional validators and organizations that will allocate significant resources to NEAR protocol.
+However, an appreciating token, driven by organic growth and the subsequent sharding dynamics, will open new opportunities for smaller operators - increasing decentralization and delegation opportunities for the stake in the network.
```
ilblackdragon: Here it's more about:
At the start, one or small number of shards require a large financial stake or large number of delegators, which will privilege professional and well known validators and organizations that will allocate significant resources to NEAR protocol and allow to come through origination stage.

As network grows, the number of seats will increase with number of shards. This will dramatically lower the barrier for stake to become validators and the seat price selection mechanics will allow for a long tail of validators to come on board.

docs/validator/economics.md
```diff
@@ -0,0 +1,132 @@
...
+This reward is fixed, regardless of the stake:
+- A relatively small stake of 100 million (10% of 1 billion) will generate a nominal return of 50% (50 million NEAR for a stake of 100 million), thus attracting more staking.
+- On the other hand, a large stake of 800 million (80% of 1 billion) will return 6.25% in rewards (50 million NEAR for a stake of 800 million), which may become too low to lock so many tokens.
+
+<blockquote class="info">
+    <strong>did you know?</strong><br><br>
+    The protocol automatically re-stakes your rewards at the end of each epoch, so you may expect up to 5% of additional stake every year.
+</blockquote>
+
+**Median of the stake across validators**
```
ilblackdragon: so, to me median is a confusing metric.

Because you can have: [1,000,000, 1, 1, 1, 1] -> the first proposal will take all 100 seats, even though median is 1.

I think we should show here an example, with proposals, how to find the seat price and number of seats. Also describe how rewards are still proportional to the stake vs # seats.

Something like, given proposals (with rollovers from previous epoch):
| Validator | Stake | # Seats |
| - | - | - |
V1 |   1,000,000    | 48
V2 |      500,000   | 24
V3 |      300,000   | 14 
V4 |      300,000 | 14
--- | seat price is 20,500 |  --- |
V5 |         20,000 | 0 |

Where seat price is such an integer number, such that if you divide all proposals and round down by it - the number of seats will be greater or equal than required.

docs/validator/economics.md
```diff
@@ -0,0 +1,132 @@
...
+
+## 3. Continuous operation
+
+Validators must keep the operations up and running by:
+1. Setting the *Service Level Objectives*
+2. Leveraging market dynamics
+
+### Setting the Service Level Objectives
+
+Achieving 90% of uptime is not sufficient: any new release of the node should be tested and applied without interruptions. Leveraging infrastructure-as-code usually simplifies ongoing maintenance, increasing efficiency and profitability at the cost of higher upfront investments.
```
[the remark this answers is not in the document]
ailisp: Going to move devnet and betanet draft on notion here this week as they mostly finalized and doesn't change a lot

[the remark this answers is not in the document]
jimmy3dita: I'm a bit confused.
What if the sentence says "Achieving >90% or uptime"?

(note the `>` symbol)

docs/validator/economics.md
```diff
@@ -0,0 +1,132 @@
...
+On top of ordinary maintenance, your operations will require an incident/response playbook, to promptly resolve any infrastructure failure, as required by the `MINIMUM_THRESHOLD`.
+
+
+### Leverage market dynamics
+
+NEAR validators are subject to market and game-theory dynamics, relating to:
+1. The total of NEAR tokens at stake
+2. The median of the stake across validators
```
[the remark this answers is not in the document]
jimmy3dita: Do you have a link I can use to fix this part?

docs/validator/economics.md
jimmy3dita: It's there, but agree it falls easily out of scope. Removed

docs/validator/economics.md
jimmy3dita: Less is more - we can always add something in the future. Removed =)

docs/validator/economics.md
jimmy3dita: My problem is explaining how `argmax` is applied to this process. Working on it 🤔

docs/validator/economics.sublime-workspace
```diff
@@ -0,0 +1,405 @@
+{
```
ailisp: This file seems to be added by mistake

jimmy3dita: Wondering how did it got there 🤔

docs/validator/economics.md
```diff
@@ -0,0 +1,136 @@
+---
+id: economics
+title: The Economics of a Validator
+sidebar_label: Understanding Economics
+---
+## Overview
```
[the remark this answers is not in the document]
jimmy3dita: It's the common standard for the docs, I suppose H1 is reserved

docs/validator/economics.md
```diff
@@ -0,0 +1,136 @@
...
+| - | - | - |
+| Total $NEAR at stake | Total seats | $NEAR per-seat |
+| 400 million | 100 | 4 million |
+| 400 million | 400 | 1 million |
+| 400 million | 800 | 0.25 million |
+
+More details can be found in the validators section of the [Economics White Paper](https://nearprotocol.com/papers/economics-in-sharded-blockchain/#validators)
+
+**Protocol fees and inflation**
+NEAR Protocol's rewards follow the formula `total_supply * 0.05/730` of NEAR tokens at the end of every `epoch` (producing up to 5% of new tokens per year). However, the protocol *burns* `tx_fees + state_rent` at the end of each block. Therefore, while rewards increase `total_supply` of $NEAR tokens, the fees *reduce* it.
```
bowenwang1996, line 117: We no longer have state rent

docs/validator/economics.md
```diff
@@ -0,0 +1,136 @@
...
+As mentioned above, there's no slashing for downtime. However, any new release of the node should be locally tested and deployed with minimum interruptions, to avoid the risks of getting *kicked out*. Allocating resources on infrastructure-as-code can increase efficiency and profitability, making ongoing maintenance easier.
+On top of ordinary maintenance, your operations will require an incident/response playbook, to promptly resolve any infrastructure failure or emergency updates.
+
+
+### Understand market dynamics
+
+NEAR validators are subject to market and game-theory dynamics, relating to:
+1. The total of NEAR tokens at stake
+2. The distribution of the stake across validators
+3. Protocol fees and inflation
```
bowenwang1996, line 89: what does "protocol fees" mean? Does this refer to transaction fees?

docs/validator/economics.md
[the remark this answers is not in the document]
amgando: H1 is automatically inserted by this line

```
title: The Economics of a Validator
```

docs/validator/economics.md
jimmy3dita: corrected. Yes it's transaction fees - I agree it's more clear

Issue #275, opened 2020-03-23
[TUTORIAL] zero to hero tutorial needs rework to remove studio


janedegtiareva:
This tutorial relies heavily on studio and needs significant rework as we deprecate studio in favor of gitpod.

janedegtiareva:
We no longer reference studio from zero to hero tutorial.

Issue #274, opened 2020-03-23
[TUTORIAL] test your smart contracts tutorial needs rework to remove studio


janedegtiareva:
This tutorial relies heavily on studio which is being deprecated and also near place which is not up to date.

Issue #273, opened 2020-03-23
how-to-write-contracts-that-talk-to-each-other needs rework to remove studio


janedegtiareva:
This tutorial relies heavily on studio, and we need to rework it to align with the move to gitpod.

amgando:
this is the related docs page: https://docs.near.org/docs/tutorials/how-to-write-contracts-that-talk-to-each-other

amgando:
start by cloning this repo: https://github.com/near-examples/token-contract-as

potatodepaulo:
In order to accomplish this:
* [ ] Read through the existing tutorial to understand the desired end result.
* [ ] Remove all instances of "NEAR studio" and replace them with generic IDE references or references to GitPod.
* [ ] Change the flow to rely on the example linked above, which can also be seen [here](https://examples.near.org/token-contract-as)
* [ ] Complete the tutorial without NEAR Studio by copying and pasting the example code.

PR #278, opened 2020-03-25
add notice re Gitpod replacing NEAR Studio

amgando:


amgando:
this PR addresses a quick triage related to #273 , #274 , #275 which all require more work

PR #285, opened 2020-03-31
docs snippets automation and including code snippets in tests

janedegtiareva:


website/replacesnippets.js
```diff
@@ -0,0 +1,26 @@
...
+const options = {
+
+    //Glob(s)
+    files: [
+      'tests/*',
+      '../docs/**/*'
+    ],
+
+    //Replacement to make (string or regex)
+    from: "<snippet id='shell-install-command'/>",
```
janedegtiareva: this will be read from a data file (probably simple xml)

website/tests/testcli.sh
```diff
@@ -0,0 +1 @@
+<snippet id='shell-install-command'/>
```
janedegtiareva: This runs in CI already (the build failed for other reasons. this part passed.)

chadoh: What is this testing?

janedegtiareva: in this particular case, that whatever we are using to install shell works. This would be followed by actual shell usage when this goes out of the WIP stage (e.g. invoke near state test.near as the next line)

mikedotexe: An immediate thought is we're going to have commands like those here:
https://docs.nearprotocol.com/docs/local-setup/local-dev-testnet#4-create-an-account-for-your-contract
`near create_account <yourcontractname> --masterAccount=<login> --initialBalance <initalbalance>`
in multiple places that would benefit from being snippets, but shouldn't be tested. wonder if we should add an attribute `notest` or something.
But this comment is not a blocking one. I think this is great.

docs/development/near-clitool.md
```diff
@@ -22,7 +22,7 @@ You use NEAR Shell to connect to the NEAR platform from the terminal to create a
 To install NEAR Shell globally (recommended)
 
 ```bash
-npm install -g near-shell
+<snippet id='shell-install-command'/>
```
chadoh: This syntax feels strange to me. I also don't like that we have three places to look to fully understand what's happening here:

1. this file (and anywhere else that might use this syntax)
2. the place where we define that this snippet should expand to `npm install -g near-shell`
3. the place where we test that `npm install -g near-shell` behaves as we expect

If we are building the processor for these snippets ourselves, we have a lot of liberty with this syntax. How would you feel about something like this?

```markdown
To install NEAR Shell globally (recommended):

    npm install -g near-shell

<!-- TEST-SNIPPET
execSync('npm install -g near-shell')
const output = execSync('near state test.near')
assertEqual(output, 'whatever')
-->
```

Note here that I'm not automating consistency between the first line of the test and the actual command text, but since they're right next to each other it should be straightforward for humans to notice if they become out-of-sync. Also, if we didn't run this through a preprocessor, it would still look correct, since the test code is in a comment.

Then our preprocessor can look for comment blocks starting with `TEST-SNIPPET` and run them with `jest` or whatever, and include things like `execSync` so that our test snippets don't get cluttered. (I'm completely unconvinced of the `TEST-SNIPPET` identifier, or using jest, or using `execSync`. Feel free to suggest alternatives to all of these.)

janedegtiareva: I think this is a lot harder to do. Our tests are going to be using diff technologies depending on what we are testing (e.g. near shell testing is best via a cli script, assembly script contract code is best using the unit testing tools we use in the examples, etc. It's very easy to do if we have explicit tests that reference snippets but very hard the other way around.

janedegtiareva: Another feature we'll lose with defining tests in the .md file: we can no longer de-duplicate some content. Specifically for near-shell installation line, it features in several pages and will continue to be duplicated like this, except now possibly with dupe tests as well. If we continue with this PR, all the places that need a "how to install shell" instruction can just reference this content which is defined in one place.

website/tests/testcli.sh
janedegtiareva: Tests are pretty much optional right now (i.e. someone needs to explicitly make a test for something to be tested), so what you are suggesting pretty much is the default situation anyways.

docs/development/near-clitool.md
chadoh: Ah, gotcha, that makes sense to me. Thanks for explaining and expanding the example!

PR #317, opened 2020-04-23
add links to new API documentation

amgando:
update sidebar nav in API section of docs

image


amgando:
@chadoh does this structure make sense?  nothing on the pages yet but presumably those would include links to `near-examples` using the relevant API, unit tests, workshops, etc to get people started quickly.

chadoh:
Yes, makes sense to me!

PR #323, opened 2020-04-28
 refactor(rpc): Move the query parameters documentation to the RPC reference page

frol:


amgando:
i think we can remove this stuff now too, right?
image

docs/interaction/rpc.md
```diff
@@ -183,6 +183,237 @@ near call dev-jdvw47f9j setResponse --args '{"apiResponse": "hello"}' --accountI
...
+    "block_hash": "4MkU8Xy82kD875NwXo7gh5wrRbpLrE5gbJ38JHtpp7m1"
+  },
+  "id": "dontcare"
+}
+```
+
+<blockquote class="warning">
+<strong>heads up</strong><br><br>
+
+`[48]` is an array of bytes, to be specific it is an ASCII code of `0`, near-sdk-rs and near-as return a JSON-serialized results.
```
amgando, line 407: should be
> near-sdk-rs and near-**sdk**-as

but even better to use backticks since they're special names
> `near-sdk-rs` and `near-sdk-as`

docs/interaction/rpc.md
```diff
@@ -372,25 +596,17 @@ http post https://rpc.nearprotocol.com/ jsonrpc=2.0 method=query params:='["call
...
-      111,
-      108,
-      46,
-      110,
-      101,
-      97,
-      114,
-      34
-    ]
+    "result": "ImZyb2wubmVhciI="
```
amgando, line 599: wow, you even updated the deprecated results.  strong move

PR #326, opened 2020-04-29
feat: Remove stale artifacts from /docs

chadoh:
Our old workflow:

* in the `nearlib`, `near-bindgen`, & `near-runtime-ts` projcets, we would build docs
* we would copy the generated markdown files into this repository's `/docs` folder
* when we build the user-facing website here, these docs would be included

These three libraries have since been renamed to `near-api-js`, `near-sdk-rs`, and `near-sdk-as`. Furthermore, we have stopped using the above workflow, opting instead to link from this site to external API Reference documentation for each of the above projects.

So we no longer need these directories.


amgando:
@chadoh just cleaned up a few things and fixed some broken links after the removal of API docs.  CI should pass now.  also renamed a bunch of `near-runtime-ts` to `near-sdk-as` (long overdue) and added a couple ignores to the `.travis.yml` file and `test-links.sh` to keep them in sync.  should be good to go.

@janedegtiareva small note re: code snippet replacement -- it makes docs workflow less convenient if we don't clean up the modified snippets in the `.md` files after the replacement process finishes (like, i have to `git reset` a bunch of files to keep from committing the replacements while testing locally)

amgando:
the build is hanging on the new developer metrics opt-in

not sure why we even have this?  @janedegtiareva or @chadoh do either of you know the origin of this file?  i don't recall ever needing NEAR Shell as part of docs. 

can i just remove this `website/tests/testcli.sh` file?

```
# make sure that the cli installs correctly
npm install -g near-shell

NODE_ENV=ci near state test.near

# test app with snippets
npx create-near-app assemblyscript-react
cd assemblyscript-react
#yarn TODO: investigate why this fails on build
```

currently the build is hanging here

```
$ tests/testcli.sh
/home/travis/.nvm/versions/node/v12.16.3/bin/near -> /home/travis/.nvm/versions/node/v12.16.3/lib/node_modules/near-shell/bin/near
+ near-shell@0.21.0
added 203 packages from 163 contributors in 6.984s
Using options: {
  networkId: 'shared-test',
  nodeUrl: 'http://shared-test.nearprotocol.com:3030',
  contractName: undefined,
  masterAccount: 'test.near',
  accountId: 'test.near',
  initialBalance: null
}
prove developer experience. We will never send private information. 
We only collect which commands are run via an anonymous identifier. 
Would you like to opt in (y/n)?
```

amgando:
ok, looking at the file history it seems like this `website/testcli.sh` file was added as part of the code snippet replacement and `.travis.yml` is triggering the file.  still not sure how it helps with docs but i think @janedegtiareva has more context on this.  will wait to hear back but commenting the line out of `.travis.yml` now to get the build passing.

PR #329, opened 2020-05-02
feat(RPC): added a useful comment about the finality

frol:


docs/api/rpc-params.md
```diff
@@ -24,7 +24,11 @@ The following identifiers are used throughout the code samples below.
 
 *`block_id` and `finality` are mutually exclusive parameters -- use one or the other, but not both.*
 
-*At time of writing, `finality` is only available on StageNet.*
+<blockquote class="warning">
+<strong>heads up</strong><br><br>
+
+The blocks that did not reach the `final` finality status (i.e. `optimistic` and `near-final`) may disapear if re-org after a fork happens. You can only count on the finalized blocks, but in some cases it is good to have the access to the latest blocks.
```
bowenwang1996, line 30: What do you mean by "may disappear"? Even if a reorg happens the blocks won't "disappear". They are still there. Also for `near-final` a reorg would mean that someone gets slashed.

frol: If there was a fork, the block with height X will be either different or non-existent.

Feel free to send an update right into this branch.

amgando: @bowenwang1996 and @frol how should we handle this?  i'm not clear enough on these details to integrate your comments.

bowenwang1996: It is fine to leave it as is.

frol: @amgando I think we can iterate from here in the future once we have a better idea on how to present different states the chain can get into. E.g. I am not sure we have a good explanation of the forks (we have DoomSlug whitepaper, but it is too detailed).

Let's merge this now.

PR #354, opened 2020-05-22
change NEAR Shell URL

amgando:


chadoh:
@amgando test failures seem legit

amgando:
yep, forgot to change the id in the file.  also renamed it

PR #365, opened 2020-05-29
update hardware requirements

### LINKED ISSUE #364: [BUG] Hardware requirement for running a node is off
In this page https://docs.near.org/docs/roles/validator/hardware we claim that 50GB of disk space is enough, which is not true. I suggest the following:
* [x] Change it to "at least 100GB" and update the cost correspondingly.
* [x] Remove "Windows" in the table as we don't really support it (and has no plan to do so in the foreseeable future).
* [ ] There is no requirement on whether the disk is SSD or HDD. We should not specify SSD here (could be misleading).

amgando:
closes #364 

@bowenwang1996 do these changes look reasonable?

see here: https://docs-pr-365.onrender.com/docs/roles/validator/hardware

### DIFF
--- a/docs/roles/validator/hardware.md
+++ b/docs/roles/validator/hardware.md
@@ -16,11 +16,11 @@ This page covers the minimum and recommended hardware requirements for engaging 
 
 Estimated monthly costs depending on operating system:
 
-| Cloud Provider | Machine Size    | Linux | Windows |
-|----------------|-----------------|-------|---------|
-| AWS            | c5.large        | $25   | $100    |
-| GCP            | n1-standard-2   | $50   | $150    |
-| Azure          | Standard_F2s_v2 | $90   | $150    |
+| Cloud Provider | Machine Size    | Linux |
+| -------------- | --------------- | ----- |
+| AWS            | c5.large        | $25   |
+| GCP            | n1-standard-2   | $50   |
+| Azure          | Standard_F2s_v2 | $90   |
 
 ### Track all shards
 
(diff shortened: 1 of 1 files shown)


docs/roles/validator/hardware.md
```diff
@@ -4,46 +4,52 @@ title: Hardware Requirements
...
+This page covers the minimum and recommended hardware requirements for engaging with the NEAR platform as an application developer, validator, integrator or core contributor.
 
 ## Machine Specifications
 
 ### Validator
 
 - **CPU:** 2-Core (4-Thread) Intel i7/Xeon equivalent
 - **RAM:** 4GB DDR4
-- **SSD:** 50GB
+- **SSD:** 100GB
```
bowenwang1996, line 15: Doesn't have to be SSD

docs/roles/validator/hardware.md
```diff
@@ -4,46 +4,52 @@ title: Hardware Requirements
...
 
 ### Track all shards
 
-Currently, if you want to ensure full security immediately after finality gadget is done, the easiest way is to run node that tracks all shards.
+Currently, if you want to ensure full security immediately after the finality gadget is done, the easiest way is to run a node that tracks all shards.
 
 - **CPU:** 8-Core (16-Thread) Intel i7/Xeon or equivalent
 - **RAM:** 16GB DDR4
-- **SSD:** 120GB
+- **SSD:** 200GB
```
bowenwang1996, line 31: For a single shard network this should be the same as running a validator

amgando:
@bowenwang1996 applied changes.  assuming good to go: merging and closing

PR #382, opened 2020-06-05
Add external link feature

### LINKED ISSUE #182: [META] external links should be distinct from internal links
**Describe the bug**
Today it's difficult to know whether clicking a link will take me out of the documentation or deeper into the documentation

**Expected behavior**
It would be better to have some visual indication of external links

**Screenshots**
several external links on this page
image

example of external link indicator
image

**:pencil: Location**
Here's one page where this would make a difference to readers
https://docs.nearprotocol.com/docs/contribution/contribution-overview

icerove:
fix #182


icerove:
@amgando CI is not passed. Do you know whether it is okay or how can I make it pass?

amgando:
@icerove this isn't a good way to update links because now we have to remember to add this `img` tag to each external URL.  it would be better to do one of the following:

- (EASY but a little SLOPPY) either add a CSS class called something like `link-external` that injects the image to the `:after` pseudoclass (i think that's what it's called)
- (HARDER but CLEAN) or add some pre-processing step to the website build process that checks link URLs and automatically adds some external link icon if they're not on the docs domain

it's also possible that this feature is already included in Docusaurus v2 so you may want to check that out.

in general, these meta data edits are low priority so if you don't have time for this but still want to contribute to the docs, there are over 50 (!!) issues that could use your help 😉

amgando:
for the record, I closed this without merging because the solution is not acceptable

PR #385, opened 2020-06-08
Translate validator setup document into korean language

NAKsir-melody:
Contribution for koreans

Signed-off-by: sigmoid <jiyun.kim@dsrvlabs.com>


amgando:
this is great, thank you @NAKsir-melody -- we don't have i18n enabled at the moment so i will just include a link to your translated page from the top of the original page before merging.

PR #390, opened 2020-06-09
minor cleanup

amgando:
@navneetlakra here's the PR that I used to make some edits to your contribution.

thanks again for doing that!  would be great to meet _before_ you start the work next time so we can make sure to minimize changes and make the best use of your time.  

here's the page
https://docs-pr-390.onrender.com/docs/local-setup/running-testnet


navneetlakra:
Sure, Sherif. Will discuss before I start on anything else.

PR #400, opened 2020-06-24
remove reference to old zero to hero tutorial from intro

mikedotexe:
Noticed that while we had removed the Zero to Hero guide because it's outdated and will cause issues, there was a reference in the introduction. Removed that.


mikedotexe:
You'll note that the broken link checker failed because of a 429 "too many requests" on https://twitter.com/argenthq
I made sure it still exists and will merge

PR #403, opened 2020-07-02
fix: create-near-app now defaults to Vanilla

chadoh:
`create-near-app` is being updated to default to vanilla JS in near/create-near-app#387

There have been some other changes lately, too, and this updates the docs to match

Additionally, I doubt it makes sense to have an entire dedicated page for `create-near-app` in the docs, so I've removed it. This sets us up to have duplicated docs here and in the `create-near-app` README. This moves us closer to a single source of truth and makes the docs easier to maintain.


chadoh:
@amgando I sort of liked that `create-near-app` was mentioned in the `basics` section, but I didn't love that it had a whole page instead of a simple link to the repo where you can learn more from its own README. However, I'm not sure where such a link fits into the Basics section. Given that we already link to `create-near-app` in many other spots in the docs, I thought maybe we didn't need it in Basics at all – what do you think?

chadoh:
marking as draft so we don't accidentally merge before https://github.com/near/create-near-app/pull/387

mikedotexe:
@dependabot rebase

mikedotexe:
> marking as draft so we don't accidentally merge before [near/create-near-app#387](https://github.com/near/create-near-app/pull/387)

marking as ready for review as that one is merged

mikedotexe:
@dependabot rebase

PR #405, opened 2020-07-06
Temporarily take down node upgrade page

bowenwang1996:
Take down the node upgrade page for now as the zero-downtime upgrade doesn't quite work.


jimmy3dita:
This will close also #402 as maintenance.md would stay.

PR #409, opened 2020-07-15
feat: add docs/concepts/storage with pricing info

chadoh:
Partially addresses near/devx#235

People need to know how much it's going to cost to run a smart contract on NEAR. This new page explains the what, why, and how of state staking.

### DIFF
--- a/docs/concepts/storage.md
+++ b/docs/concepts/storage.md
@@ -1,0 +1,141 @@
+---
+id: storage
+title: Storage Staking
+sidebar_label: Storage Staking
+---
+
+When you deploy a smart contract to NEAR, you pay for the storage that this contract requires using a mechanism called storage staking. (You may have heard that NEAR charges rent for storage – this has not been true [since March 2020](https://github.com/nearprotocol/nearcore/pull/2272)).
+
+In storage staking (sometimes called _state_ staking), the account that owns a smart contract must stake (or lock) tokens according to the amount of data stored in that smart contract, effectively reducing the balance of the contract's account.
+
+<blockquote class="info">
+<strong>Coming from Ethereum?</strong><br><br>
+
+If you’re familiar with Ethereum’s pricing model, you may know that, like NEAR, the protocol charges a fee (called "gas") for each transaction. Unlike NEAR, Ethereum's gas fee accounts for the amount of data stored via that transaction. This essentially means that anyone can pay once to store permanent data on-chain. This is a poor economic design for at least two reasons: 1. The people running the network (miners, in the case of Ethereum 1) are not appropriately incentivized to store large amounts of data, since a gas fee far charged in the distant past can increase storage costs forever, and 2. The users of a smart contract are charged for the data they send to store in it, rather than charging the owner of the smart contract.
+</blockquote>
+
+## How does NEAR's design align incentives?
+
+Storage-staked tokens are unavailable for other uses, such as validation staking. This increases the yield that validators will receive. Learn more in [the economics whitepaper](https://near.org/papers/economics-in-sharded-blockchain/).
+
+## When do tokens get staked?
+
+On each incoming transaction that adds data.
+
+Let's walk through an example:
+
+1. You launch [a guest book app](https://examples.near.org/guest-book), deploying your app's smart contract to the account `example.near`
+2. Visitors to your app can add messages to the guest book. This means your users will, [by default](https://docs.near.org/docs/concepts/gas#what-about-prepaid-gas), pay a small gas fee to send their message to your contract.
+3. When such a call comes in, NEAR will check that `example.near` has a large enough balance that it can stake an amount to cover the new storage needs. If it does not, the transaction will fail.
+
+## The "million cheap data additions" attack
+
+Note that this can create an attack surface. To continue the example above, if sending data to your guest book costs users close to nothing while costing the contract owner significantly more, then a malicious user can exploit the imbalance to make maintaining the contract prohibitively expensive.
+
+Take care, then, when designing your smart contracts to ensure that such attacks cost potential attackers more than it would be worth.
+
+## btw, you can remove data to unstake some tokens
+
+People familiar with the "immutable data" narrative about blockchains find this surprising. While it's true that an _indexing node_ will keep all data forever, _validating nodes_ (that is, the nodes run by most validators in the network) do not. Smart contracts can provide ways to delete data, and this data will be purged from most nodes in the network within a few [epochs](https://docs.near.org/docs/concepts/epoch).
+
+Note that a call to your smart contract to remove data has an associated gas fee. Given NEAR's gas limit, this creates an upper limit on how much data can be deleted in a single transaction.
+
+## How much does it cost?
+
+Storage staking is priced in an amount set by the network, which was initialized to **[9.09E19 yoctoNEAR per byte](https://github.com/nearprotocol/nearcore/blob/master/neard/res/genesis_config.json#L38)**, or **just over 10kb per NEAR token (Ⓝ)**
+
+This value may change in the future. NEAR's JSON RPC API provides [a way to query this initial setting](https://docs.near.org/docs/api/rpc-experimental#genesis-config), but does not yet provide a way to query the "live" configuration value. Before it changes, this document will be updated to include information about how to query the live version.
+
+## Example cost breakdown
+
+Let's walk through an example.
+
+A [non-fungible token](https://github.com/nearprotocol/NEPs/pull/4) is unique, which means each token has its own ID. The contract must store a mapping from token IDs to owners' account ID.
+
+If such an NFT is used to track **1 million** tokens, how much storage will be required for the token-ID-to-owner mapping? And how many tokens will need to be staked for that storage?
+
+Using [this basic AssemblyScript implementation](https://github.com/near-examples/NFT/tree/master/contracts/assemblyscript/nep4-basic) as inspiration, let's calculate the storage needs when using a [`PersistentMap`](https://near.github.io/near-sdk-as/classes/_assembly_sdk_collections_persistentmap_.persistentmap.html) from `near-sdk-as`. While its specific implementation may change in the future, at the time of writing `near-sdk-as` stored data as UTF-8 strings. We'll assume this below.
+
+Here's our `PersistentMap`:
+
+```ts
+type AccountId = string
+type TokenId = u64
+const tokenToOwner = new PersistentMap<TokenId, AccountId>('t2o')
+```
+
+Behind the scenes, all data stored on the NEAR blockchain is saved in a key-value database. That `'t2o'` variable that's passed to `PersistentMap` helps it keep track of all its values. If your account `example.near` owns token with ID `0`, then at the time of writing, here's the data that would get saved to the key-value database:
+
+* key: `t2o::0`
+* value: `example.near`
+
+So for 1 million tokens, here are all the things we need to add up and multiply by 1 million:
+
+1. The prefix, `t2o`, will be serialized as three bytes in UTF-8, and the two colons will add another two. That's 5 bytes.
+2. For an implementation where `TokenId` auto-increments, the values will be between `0` and `999999`, which makes the average length 5 bytes.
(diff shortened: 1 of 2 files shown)


docs/concepts/storage.md
```diff
@@ -0,0 +1,73 @@
...
+
+People familiar with the "immutable data" narrative about blockchains find this surprising. While it's true that an _indexing node_ will keep all data forever, _validating nodes_ (that is, the nodes run by most validators in the network) do not. Smart contracts can provide ways to delete data, and this data will be purged from most nodes in the network within a few [epochs](https://docs.near.org/docs/concepts/epoch).
+
+Note that a call to your smart contract to remove data has an associated gas fee. Also, since deletions takes time, there is an upper limit on the amount of data you can remove in a single transaction, as your transaction must be able to fit in the one-second block time.
+
+## How much does it cost?
+
+State staking is priced in an amount set by the network, which was initialized to **[9.09E19 yoctoNEAR per byte](https://github.com/nearprotocol/nearcore/blob/master/neard/res/genesis_config.json#L38)** and could be lowered (but not increased) in the future. NEAR's JSON RPC API provides [a way to query this initial setting](https://docs.near.org/docs/api/rpc-experimental#genesis-config), but does not yet provide a way to query the "live" configuration value.
```
chadoh, line 41: @bowenwang1996 is it true that `storage_amount_per_byte` will never be increased?

bowenwang1996: No

chadoh: It could be increased? What would happen to all the contracts with insufficient free balance to stake for the higher cost?

bowenwang1996: It depends on how we upgrade. For example we can make it such that the contract will be removed after some amount of time if no one funds it. Alternatively we can say that we will only allow transfer to this account until if has sufficient tokens to pay for storage.

docs/concepts/storage.md
```diff
@@ -0,0 +1,92 @@
...
+
+Note that this can create an attack surface. To continue the example above, if naming an escrow costs a token holder close to nothing while costing the contract owner significantly more, then a malicious user can exploit the imbalance to make maintaining the contract prohibitively expensive.
+
+Take care, then, when designing your smart contracts to ensure that such attacks cost potential attackers more than it would be worth.
+
+## btw, you can remove data to unstake some tokens
+
+People familiar with the "immutable data" narrative about blockchains find this surprising. While it's true that an _indexing node_ will keep all data forever, _validating nodes_ (that is, the nodes run by most validators in the network) do not. Smart contracts can provide ways to delete data, and this data will be purged from most nodes in the network within a few [epochs](https://docs.near.org/docs/concepts/epoch).
+
+Note that a call to your smart contract to remove data has an associated gas fee. Also, since deletions takes time, there is an upper limit on the amount of data you can remove in a single transaction, as your transaction must be able to fit in the one-second block time.
```
mikedotexe, line 37: …in a one-second block time and due to computational limits per block. (don't know the source)

bowenwang1996: block time is not really one second. It is better to talk about gas limit here.

docs/concepts/storage.md
```diff
@@ -0,0 +1,97 @@
...
+
+Storage staking is priced in an amount set by the network, which was initialized to **[9.09E19 yoctoNEAR per byte](https://github.com/nearprotocol/nearcore/blob/master/neard/res/genesis_config.json#L38)** and could be lowered (but not increased) in the future. NEAR's JSON RPC API provides [a way to query this initial setting](https://docs.near.org/docs/api/rpc-experimental#genesis-config), but does not yet provide a way to query the "live" configuration value.
+
+Someday this value will probably change, and when it does this document will be updated to include information about how to query the live version.
+
+## Example cost breakdown
+
+Let's walk through an example.
+
+A non-fungible token is one where each token is unique, which means each token has an ID. Amongst other things, such a contract must store a mapping from token IDs to owners' account addresses.
```
willemneal, line 49:
```suggestion
A non-fungible token is unique, which means each token has its own ID. The contract must store a mapping from token IDs to owners' account ID.
```

docs/concepts/storage.md
```diff
@@ -0,0 +1,97 @@
...
+
+People familiar with the "immutable data" narrative about blockchains find this surprising. While it's true that an _indexing node_ will keep all data forever, _validating nodes_ (that is, the nodes run by most validators in the network) do not. Smart contracts can provide ways to delete data, and this data will be purged from most nodes in the network within a few [epochs](https://docs.near.org/docs/concepts/epoch).
+
+Note that a call to your smart contract to remove data has an associated gas fee. Also, since deletions takes time, there is an upper limit on the amount of data you can remove in a single transaction, as your transaction must be able to fit in the one-second block time.
+
+## How much does it cost?
+
+Storage staking is priced in an amount set by the network, which was initialized to **[9.09E19 yoctoNEAR per byte](https://github.com/nearprotocol/nearcore/blob/master/neard/res/genesis_config.json#L38)** and could be lowered (but not increased) in the future. NEAR's JSON RPC API provides [a way to query this initial setting](https://docs.near.org/docs/api/rpc-experimental#genesis-config), but does not yet provide a way to query the "live" configuration value.
```
willemneal, line 41: Might be good to translate the cost to bytes per NEAR.

docs/concepts/storage.md
```diff
@@ -0,0 +1,97 @@
...
+
+Someday this value will probably change, and when it does this document will be updated to include information about how to query the live version.
+
+## Example cost breakdown
+
+Let's walk through an example.
+
+A non-fungible token is one where each token is unique, which means each token has an ID. Amongst other things, such a contract must store a mapping from token IDs to owners' account addresses.
+
+If such an NFT is used to track **1 million** tokens, how much storage will be required, just for that token-ID-to-owner mapping? And how many tokens will need to be staked for that storage?
```
willemneal, line 51:
```suggestion
If such an NFT is used to track **1 million** tokens, how much storage will be required for the token-ID-to-owner mapping? And how many tokens will need to be staked for that storage?
```

docs/concepts/storage.md
```diff
@@ -0,0 +1,97 @@
...
+2. For an implementation where `TokenId` auto-increments, the values will be between `0` and `999999`, which makes the average length 5 bytes.
+3. Let's assume well-formed NEAR `AccountId`s, and let's guess that NEAR Account IDs follow the approximate pattern of domain names, which [average about 10 characters](https://www.domainregistration.com.au/news/2013/1301-domain-length.php), plus a top-level name like `.near`. So a reasonable average to expect might be about 15 characters; let's keep our estimate pessimistic and say 25. This will equal 25 bytes, since NEAR account IDs must use characters from the ASCII set.
+
+So:
+
+    1_000_000 * (5 + 5 + 25)
+
+35 million bytes. Multiplying by 9.09E19 yoctoNEAR per byte, we find that the `tokenToOwner` mapping will require staking 3.1815E27 yoctoNEAR, or Ⓝ3,181.5
+
+Note that you can get this down to Ⓝ2,999.7 just by changing the prefix from `t2o` to a single character
```
willemneal, line 78: You can get rid of the whole prefix too, so -3 more bytes.

mikedotexe:
approved with trust! great stuff

docs/concepts/storage.md
```diff
@@ -0,0 +1,97 @@
+---
+id: storage
+title: Storage Staking
+sidebar_label: Storage Staking
+---
+
+If you’re familiar with Ethereum’s pricing model, you may know that the protocol charges a fee (called "gas") once for a transaction, and this transaction takes into account the amount of data stored via that transaction. This essentially means that anyone can pay once to store permanent data on-chain. This is a poor economic design for at least two reasons: 1. The people running the network (miners, in the case of Ethereum 1) are not appropriately incentivized to store large amounts of data, since a gas fee far charged in the distant past can increase storage costs forever, and 2. The users of a smart contract are charged for the data they send to store in it, rather than charging the owner of the smart contract.
```
janedegtiareva, line 7: What's the target audience? Is it possible to explain to a dev who is not familiar with Ethereum? I think it's better to start with explaining what NEAR does and then say how it's different from Ehtereum.

docs/concepts/storage.md
```diff
@@ -0,0 +1,97 @@
...
+
+NEAR addresses this economic misalignment using a mechanism called storage staking (you may have heard that NEAR charges rent for storage – this has not been true [since March 2020](https://github.com/nearprotocol/nearcore/pull/2272)). In storage staking (sometimes called _state_ staking), the account that owns a smart contract must stake (or lock) tokens according to the amount of data stored in that smart contract, effectively reducing the balance of the contract's account.
```
janedegtiareva, line 9: If we start with this paragraph, this doc will be much more friendly for a non-blockchain but technically solid reader.

docs/concepts/storage.md
```diff
@@ -0,0 +1,97 @@
...
+
+## How does this align incentives?
+
+Storage-staked tokens are unavailable for other uses such as validation staking. This increases the yield that validators will receive. Learn more in [the economics whitepaper](https://near.org/papers/economics-in-sharded-blockchain/).
```
janedegtiareva, line 13: grammar nit: comma missing before "such as validation staking"

docs/concepts/storage.md
```diff
@@ -0,0 +1,97 @@
...
+
+## When do tokens get staked?
+
+On each incoming transaction that adds data.
+
+Let's walk through an example:
+
+1. You launch a [a custom token](https://examples.near.org/?q=fungible), deploying your token's smart contract to the account `mycoiniscoolerthanyours.near`
+2. Token holders can name an escrow – an account which can transfer tokens on their behalf. Naming an escrow will require a call to your smart contract, and your smart contract will need to store this mapping of token-owner-to-escrow-account.
+3. When this call comes in, NEAR will check that `mycoiniscoolerthanyours.near` has a large enough balance that it can stake an amount corresponding to this data addition. If it does not, the transaction will fail. Otherwise, these new tokens get locked for storage staking and the transaction completes.
```
janedegtiareva, line 23: can we use a different contract as an example? It can be confusing which tokens you are staking (NEAR vs mycoiniscoolerthanyours) Even switching to NFT will make it easier to grok.

docs/concepts/storage.md
```diff
@@ -0,0 +1,97 @@
...
+
+## The "million worthless escrows" attack
+
+Note that this can create an attack surface. To continue the example above, if naming an escrow costs a token holder close to nothing while costing the contract owner significantly more, then a malicious user can exploit the imbalance to make maintaining the contract prohibitively expensive.
+
+Take care, then, when designing your smart contracts to ensure that such attacks cost potential attackers more than it would be worth.
+
+## btw, you can remove data to unstake some tokens
+
+People familiar with the "immutable data" narrative about blockchains find this surprising. While it's true that an _indexing node_ will keep all data forever, _validating nodes_ (that is, the nodes run by most validators in the network) do not. Smart contracts can provide ways to delete data, and this data will be purged from most nodes in the network within a few [epochs](https://docs.near.org/docs/concepts/epoch).
```
janedegtiareva, line 35: possibly future improvement considerations: what is indexing node vs validating node?

docs/concepts/storage.md
janedegtiareva, line 78: +100 to providing suggestions on how to minimize costs! Are there any other tips we can come up with at this point?

chadoh: There's a whole section on this at the bottom 😄

chadoh: Yeah, I may as well mention that!

docs/concepts/storage.md
chadoh: @bowenwang1996 is the gas limit and the reason for it documented somewhere such as Nomicon?

docs/concepts/storage.md
chadoh: Still seems like something that could cause problems, even if mostly social problems. But thanks, I'll update the language.

docs/concepts/storage.md
chadoh: Great idea, thanks! Rearranged.

docs/concepts/storage.md
chadoh: This cleaned it up a lot; great suggestion

docs/concepts/storage.md
chadoh: I could not find documentation for these concepts; I only know that these are possibilities because @bowenwang1996 said so in a call. Bowen, do you know if indexing nodes are documented anywhere?

docs/concepts/storage.md
chadoh: ...which I now realize I failed to push before you reviewed! Sorry; I had it written but failed to push 😞

PR #425, opened 2020-08-04
change references of near shell to near cli

mikedotexe:
We're now using
https://www.npmjs.com/package/near-cli
and have a deprecation warning at the top of:
https://www.npmjs.com/package/near-shell

Also the Github has been renamed to:
https://github.com/near/near-cli


mikedotexe:
The 3 "broken links" were 429's "Too Many Requests", going to merge

PR #431, opened 2020-08-10
fix: storage class link update

thisisjoshford:
Links to information on the storage class were broken and this PR updates them to their new URL.


package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "scripts": {
+    "test": "cd ./website && ./test-links.sh"
```
chadoh, line 3: there's already a `package.json` file in `website/package.json` – would it make more sense to add this `test` script there?

.travis.yml
```diff
@@ -5,5 +5,5 @@ script:
   - cd website; yarn install; yarn build
   - npx serve build/near-docs &
   - npx wait-on http://localhost:5000/
-  - npx broken-link-checker -gro --exclude "localhost:3030" --exclude "https://github.com/nearprotocol/docs/tree/master/docs" --exclude "https://youtu.be" --exclude "http://near.ai/wbs" --exclude "https://www.youtube.com" --host-requests 2 http://localhost:5000 --exclude "https://rpc.testnet.near.org" --exclude "https://rpc.betanet.near.org" --exclude "https://rpc.near.org"  --exclude "https://crates.io/crates/near-sdk"  --exclude "https://github.com/nearprotocol/workshop"
+  - npx broken-link-checker -gro --exclude "localhost:3030" --exclude "https://github.com/nearprotocol/docs/tree/master/docs" --exclude "https://youtu.be" --exclude "http://near.ai/wbs" --exclude "https://www.youtube.com" --host-requests 2 http://localhost:5000 --exclude "https://rpc.testnet.near.org" --exclude "https://rpc.betanet.near.org" --exclude "https://rpc.near.org"  --exclude "https://crates.io/crates/near-sdk"  --exclude "https://github.com/nearprotocol/workshop" --exclude "https://wasmbyexample.dev"
```
chadoh, line 8: While we're here it probably makes sense to add `broken-link-checker` as a dev dependency in `website` (`npm i -D broken-link-checker`, then we can use `cd website && yarn test` here instead of relying on a "live" version of `broken-link-checker`

package.json
thisisjoshford: I did create a `test` script there as well.  This was created so one could run `yarn test` in either directory and it would work. Kind of a lazy person's script lol.  I can remove it if you like.

.travis.yml
thisisjoshford: sounds good!  I'll make the changes

thisisjoshford: made the suggested revisions and the Travis build seemed to run the `broken-link-checker` just as before, however a broken link did not cause Travis to fail; it showed a false passing. Spent some time trying to find a solution, but after my timeboxed session I was not successful and just reverted back to original code. Maybe you could take a look at my commit history and suggest a working solution?

PR #448, opened 2020-09-01
link to experimental methods

### LINKED ISSUE #443: [DOC] Document rpc for querying state changes
To understand the changes to accounts at a given block precisely, it is often necessary to query state changes through rpc. We currently don't have much documentation on how to use those endpoints (`EXPERIMENTAL_changes` and `EXPERIMENTAL_changes_in_block`).

thisisjoshford:
- reference & link to state changes in experimental methods MD
- relocate `rpc.MD` to `api` directory for consistency 

fixes #443


thisisjoshford:
How does "To view state changes as well as other features we are working on, please visit our [Experimental Methods section](/docs/api/rpc-experimental). " sound @bowenwang1996

frol:
Experimental API are feature-complete, but we do not guarantee them to be backward-compatible or exist forever. For example, we are going to remove `EXPERIMENTAL_genesis_records` API soon (https://github.com/near/near-explorer/issues/393) and `EXPERIMENTAL_genesis_config` does not have a stable structure (we used to change it quite frequently); we are committed to never do such changes for the non-EXPERIMENTAL endpoints. EXPERIMENTAL endpoints are useful for API evaluation on the real network. While the API is in EXPERIMENTAL state we collect the feedback and can iterate on the input/output structure.

P.S. I guess, we should put such a description to the docs to clarify what "EXPERIMENTAL" means.

bowenwang1996 (approved):

frol (approved):

PR #456, opened 2020-09-08
feat: add docs on transaction status, atomicity, and finality

bowenwang1996:


docs/concepts/transaction.md
```diff
@@ -50,7 +50,108 @@ There are several ways of creating `Receipts`:
...
+- `SucccessValue`.  This indicates that the receipt has been successfully executed and the value of the key
+is the return value (can only be nonempty when it is the result of a function call receipt).
+- `SuccessReceiptId`. This indicates that either a transaction has been succesfully converted to a receipt or a receipt is
+successfully processed and generated another receipt. The value of this key is the id of the newly generated receipt.
+- `Failure` This indicates that the transaction or receipt has failed.
+- `Unknown` This indicates that the transaction or receipt hasn't been processed yet.
+
+The top level `status` indicates the overall status of the transaction. For transaction that does not include function
+call actions, this means whether all the actions are successfully executed. For function call transaction, this indicates
+whether the function call has been successfully executed. However, one caveat is that the successful execution of the
```
evgenykuzyakov, line 138: The `SuccessValue` or `SuccessReceiptId` is the execution result of the last action assuming all actions before are successfully executed. If a transaction consists of 2 function calls actions, only the value of the last one is returned, while the value of the previous one is dismissed. But if the first function call fails, then the 2nd one is not executed.

docs/concepts/transaction.md
```diff
@@ -50,7 +50,108 @@ There are several ways of creating `Receipts`:
...
+function call does not necessarily mean that the receipts spawned from the function call are all successfully executed.
+For example, 
+```rust
+pub fn transfer(receiver_id: String) {
+    Promise::new(receiver_id).transfer(10);
+}
+```
+this function schedules a promise, but its return value is unrelated to the promise that it has scheduled. So even if the
+promise fails, potentially due to a lack of attached deposit, a transaction that calls this function will still have
```
evgenykuzyakov, line 147: > potentially due to a lack of attached deposit

This will fail at the outgoing receipt validation, so the TX will fail.

Better to say if the `receiver_id` doesn't exist

docs/concepts/transaction.md
```diff
@@ -50,7 +50,111 @@ There are several ways of creating `Receipts`:
...
+- `SucccessValue`.  This indicates that the receipt has been successfully executed and the value of the key
+is the return value (can only be nonempty when it is the result of a function call receipt).
+- `SuccessReceiptId`. This indicates that either a transaction has been successfully converted to a receipt or a receipt is
+successfully processed and generated another receipt. The value of this key is the id of the newly generated receipt.
+- `Failure` This indicates that the transaction or receipt has failed during execution.
+- `Unknown` This indicates that the transaction or receipt hasn't been processed yet.
+
+NOTE: for receipts, `SuccessValue` and `SuccessReceiptId` comes from the execution of the last action. The results of other
+action executions in the same receipt are not returned. However, if any action execution failed, the execution of the receipt
+stops and the failure is returned., meaning that `status` would be `Failure`.
```
evgenykuzyakov, line 138:
```suggestion
stops and the failure is returned., meaning that `status` would be `Failure`. And if the last action is not a function call and it's successfully executed, the result will be an empty `SuccessValue`
```

evgenykuzyakov (approved):

PR #460, opened 2020-09-10
removed Snippet Script

doriancrutcher:
We no longer use Snippets and Snippets Script was removed from `package.json`


thisisjoshford (approved):
Sweet! thank you!!

PR #461, opened 2020-09-10
fix seat price math and tune number of seats wording

kucharskim:
Initially spotted in https://near.org/blog/near-protocol-economics/
and then in this document. I think seat price is 21,200.

      $ bc
      1000000 + 500000 + 300000 + 300000 + 20000
      2120000
      48 + 24 + 14 + 14
      100
      2120000 / 100
      21200

Also I found wording confusing, as when one thinks about division,
there is dividend and divisor and sentence was missing explanation
what is that divisor.

With another part of the same sentene I've changed

> you will get required number of seats

to

> you will get assigned number of seats

as in my mind required number of seats is defined by the blockchain
configuration and assigned would be just in that epoch, just for that
validator.

### DIFF
--- a/docs/validator/economics.md
+++ b/docs/validator/economics.md
@@ -129,10 +129,10 @@ Let's say you observe a set of proposals (and rollovers from previous epoch):
 | V2        | 500,000              | 24      |
 | V3        | 300,000              | 14      |
 | V4        | 300,000              | 14      |
-| ---       | seat price is 20,500 | ---     |
+| ---       | seat price is 21,200 | ---     |
 | V5        | 20,000               | 0       |
 
-The *seat price* given this proposals is determined by finding such integer number that if you divide each validator's stake with rounding down (e.g. V5 20,000 / 20,500 with rounding down will be 0) - you will get required number of seats. This determines who will get their seat(s) and who's funds will be returned back to them.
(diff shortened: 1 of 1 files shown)


kucharskim:
Hopefully this makes sense, cc @bowenwang1996.

If this diff is correct I would also fix https://near.org/blog/near-protocol-economics/ article, but not sure where it is.

kucharskim:
Actually, there is still something not right with the math, because seat assignment doesn't add up:

```
1000000 / 21200
47

500000 / 21200
23

300000 / 21200
14

300000 / 21200
14

20000 / 21200
0

47 + 23 + 14 + 14
98
```

bowenwang1996:
@kucharskim Please note that, seat price **does not** equal to the total stake divided by the number of seats. You can find details on how to calculate the seat price here https://near.org/papers/economics-in-sharded-blockchain/#validators.

kucharskim:
I guess you can hold off with review, I need to dig deeper. Thanks @bowenwang1996 for pointing this out!

bowenwang1996:
@kucharskim if you think there is nothing wrong after reading how the seat price is computed, feel free to close this PR

kucharskim:
Ok, I'll close it.

PR #468, opened 2020-09-11
balance changes walkthrough

thisisjoshford:
- created a walkthrough to view `account_changes` from a token transfer using the `changes` endpoint


bowenwang1996 (approved):

docs/roles/integrator/exchange-integration.md
```diff
@@ -30,7 +30,41 @@ sidebar_label: Exchange Integration
 **Note:** NEAR requires transactions to be serialized in [Borsh](https://borsh.io/) which currently supports Rust, Javascript, & TypeScript.
 
 ## Balance Changes
-  -  Balance changes on accounts can be tracked by using our [changes endpoint](https://docs.near.org/docs/api/rpc-experimental#changes).
+Balance changes on accounts can be tracked by using our [changes endpoint](https://docs.near.org/docs/api/rpc-experimental#changes). You can test this out by sending tokens to an account using [`near-cli](/docs/development/near-cli).
+  
+  - First, make sure you are logged in.
+    ```bash
+    near login
```
bowenwang1996, line 37: you don't need run `near login`. As long as you have the key locally you are fine. `login` is for creating keys locally

thisisjoshford: I should have specified that. I've had a "heads up" in another doc I created that stated "Unless you created your account using `near-cli` you will need to login by typing the following in your terminal:" 

Should I replace it with that @bowenwang1996, or just remove it all together.

bowenwang1996: I think we should remove that and say something like make sure you have keys to your account locally, however you achieve that is not the concern of this document.

PR #469, opened 2020-09-11
updated finality documentation

thisisjoshford:
- updated `finality` example in `integrating.md`
- updated `json` result for `view_state` query example in `rpc.md`


bowenwang1996 (commented):
for finality I would say we should show an example of account balance change which shows
1) when the transaction first finishes and is not yet final, show that querying without `final` can give the balance change whereas querying with `final` displays the same result
2) when the block has been finalized, show that querying with `final` shows the balance change

thisisjoshford:
you got it!

bowenwang1996 (approved):

thisisjoshford:
Did not include aforementioned example, as a block finalizes in within 2seconds after transaction is complete. This makes querying the non-final state difficult and was decided to leave it out of this particular doc.

PR #470, opened 2020-09-12
added guide for processing transactions

thisisjoshford:
- added instructions & snippet for signing transactions
- added link to `transaction.ts` source code
- minor text edits

### DIFF
--- a/docs/roles/integrator/exchange-integration.md
+++ b/docs/roles/integrator/exchange-integration.md
@@ -28,6 +28,8 @@ sidebar_label: Exchange Integration
   ```
   - Once your transaction is constructed, you will then need to sign it by calling the `signTransactionObject` method and pass `transaction`, `signerId`, and a `networkId` (i.e. `testnet`, `betanet`, or `mainnet`)
 
+  **Note** The networkId is used only in `near-api-js` and is not included in the final on-chain transaction.
+
   ```js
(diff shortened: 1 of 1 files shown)


docs/roles/integrator/exchange-integration.md
```diff
@@ -26,6 +26,13 @@ sidebar_label: Exchange Integration
   
   const transaction = nearAPI.transactions.createTransaction(signerId, signerPublicKey, receiverId, nonceForPublicKey, actions, blockHash);
   ```
+  - Once your transaction is constructed, you will then need to sign it by calling the `signTransactionObject` method and pass `transaction`, `signerId`, and a `networkId` (i.e. `testnet`, `betanet`, or `mainnet`)
```
bowenwang1996, line 29: the network id is purely internal to near-api-js and is used to select the signer. We should not talk about it as if it is some general pattern

thisisjoshford: would just removing the (i.e. `testnet`, `betanet`...) suffice, or would you rather no mention of arguments passed?

bowenwang1996: I think we should just remove the mention of `networkId` or mention that it is specific to near-api-js

thisisjoshford: updated @bowenwang1996 :)

bowenwang1996 (approved):

PR #473, opened 2020-09-15
add transfer example for lockup and multisig

bowenwang1996:
Add examples of transfer from lockup and multisg contracts.


docs/roles/integrator/exchange-integration.md
```diff
@@ -110,6 +110,390 @@ result of the ones we provided. We expect that, at this stage, most of the contr
 we provide and therefore the potential nondeterminism in contract compilation should not be a concern. Still, we recommend
 taking the first approach whenever possible to support all transfers at once.
 
+**Example of transfer from a lockup contract**
+
+A contract `test.lockup.near` is deployed and we can check its owner by
```
evgenykuzyakov, line 115: Let's replace `test.lockup.near` with `bowen.lockup.near` to reflect a common relationships between owner or lockups, they would be like that
```suggestion
A contract `bowen.lockup.near` is deployed and we can check its owner by
```

docs/roles/integrator/exchange-integration.md
```diff
@@ -110,6 +110,390 @@ result of the ones we provided. We expect that, at this stage, most of the contr
...
+```bash
+> near view test.lockup.near get_owner_account_id
+View call: test.lockup.near.get_owner_account_id()
+'bowen'
+```
+Now we want to transfer some unlocked tokens with the following call
+```bash
+near call test.lockup.near transfer '{"amount":"100", "receiver_id": "bowen"}' --accountId=bowen
+```
```
evgenykuzyakov, line 124:
```suggestion
```bash
> near view bowen.lockup.near get_owner_account_id
View call: bowen.lockup.near.get_owner_account_id()
'bowen.near'
```
Now we want to transfer some unlocked tokens with the following call
```bash
near call bowen.lockup.near transfer '{"amount":"100", "receiver_id": "bowen.near"}' --accountId=bowen.near
```
```

File: docs/roles/integrator/exchange-integration.md [hunk 3]
```diff
@@ -110,6 +110,390 @@ result of the ones we provided. We expect that, at this stage, most of the contr
...
+    signer_id: 'multisigtest.testnet',
+    public_key: 'ed25519:BmVX32jhvEd8d8outiQdjf66GGYV3pb7kaxrKTdNisCz',
+    nonce: 5,
+    receiver_id: 'multisigtest.testnet',
+    actions: [
+      {
+        FunctionCall: {
+          method_name: 'confirm',
+          args: 'eyJyZXF1ZXN0X2lkIjo0fQ==',
+          gas: 30000000000000,
```
evgenykuzyakov, line 369: The amount of gas in the last `confirm` call has to be large enough to cover the execution logic of the multi-sig contract PLUS full amount of gas required to issue the intended call. In this case it's a `Transfer` action, which is relatively cheap.

In case of a function call from a multi-sig contract, the amount of gas has to be: `25Tgas + <AMOUNT_OF_ATTACHED_GAS>`   where `<AMOUNT_OF_ATTACHED_GAS>` is the gas attached to outgoing request.
NOTE: That the non-final `confirm` calls (e.g. the first `add_request_and_confirm`) don't require  `<AMOUNT_OF_ATTACHED_GAS>` to be attached, so `25Tgas` for them is enough.

PR #478, opened 2020-09-17
removed link to wallet

AnaisUrlichs:
@thisisjoshford I removed the link to the wallet page in the getting started section since the AccountID walk-through already covers how to use the wallet.

However, I left the page that simply links to the [NEAR wallet](https://docs.near.org/docs/quick-start/near-wallet) so that if people look specifically for the wallet in the search bar, they can still find it.

### DIFF
--- a/docs/local-setup/create-account.md
+++ b/docs/local-setup/create-account.md
@@ -4,6 +4,10 @@ title: Creating a NEAR Account in Testnet
 sidebar_label: Create Your Account
 ---
 
+## Try NEAR Wallet
+
+https://wallet.testnet.near.org
+
 Before you get started with NEAR, the first thing that you want to do is to set up a NEAR account on Testnet. This account will also be seeded with a balance so that you can use those funds to deploy your contracts, access applications, and stake tokens.
 
 Either read-on to learn more about the wallet or jump to [set up your wallet](#set-up-your-wallet).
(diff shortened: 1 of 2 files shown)


thisisjoshford:
Hey @AnaisUrlichs!  Just reviewed this PR and totally agree with your thoughts on redundancy and that we should consolidate our documentation. This solution you provided, however, does leave an orphan page that doesn't flow with the rest of our documentation. (see below)

Screen Shot 2020-09-17 at 2 35 31 PM

There are a few orphan pages that we need to clean up and would rather not add another one to them. I think what could be a more elegant solution would be to either update both the `create-account.md` and `near-wallet.md` to complement each other better _or_ update `create-account.md` to have a cleaner wallet section that will show up for searches.

AnaisUrlichs:
@thisisjoshford create Account ID covers everything; the problem is that if we name it wallet, crypto-newbies (which the get started section is targeting), will be confused. However, if we name it Account IDs or similar, people looking for the wallet will be confused.

I would rather, either
* remove the wallet page and try to update all associated links
* only provide the link to the wallet and link to the Account ID creation tutorial => in case the page is linked anywhere externally people would at least find valuable resources instead of an empty page link 

What do you think?

thisisjoshford:
This is getting completely redone shortly, so I closed this PR :)

PR #501, opened 2020-10-01
detailed walkthrough of transaction construction

### LINKED ISSUE #490: Instantiate transaction construction example
Currently we have a section on transaction construction https://docs.near.org/docs/roles/integrator/exchange-integration#constructing-transactions. However there is no example to cover the end-to-end process. We should add something like how a transfer transaction is constructed in very high detail.

### LINKED ISSUE #493: transaction signing documentation is incorrect
The documentation here https://docs.near.org/docs/roles/integrator/integrating#transaction-signing on transaction signing doesn't work because of this part
```
// WARNING: this sample won't work because Signature is not exported by near-api-js
const signature = await signer.signMessage(message, accountId, networkId);
const signedTx = new nearApi.transactions.SignedTransaction({transaction, signature: new Signature(signature.signature) });
```
has a couple issues:
1) `Signature` is exported
2) The construction of `SignedTransaction` is incorrect. See usage here https://github.com/near/near-api-js/blob/c524091b0301f6f32eb69b3b8b2d960cd4c8b819/src/transaction.ts#L230.

I suggest we combine this part with https://docs.near.org/docs/roles/integrator/exchange-integration#constructing-transactions to have a single source of truth.

thisisjoshford:
- created an end-to-end walkthrough of how to construct, sign, and send a transaction on the NEAR blockchain.
- [transaction-examples](https://github.com/near-examples/transaction-examples) repository was created for this tutorial

fixes #490 
fixes #493

### DIFF
--- a/docs/tutorials/create-transactions.md
+++ b/docs/tutorials/create-transactions.md
@@ -1,0 +1,272 @@
+---
+id: create-transactions
+title: Create Transactions
+sidebar_label: Create Transactions
+---
+To construct & process transactions you will need our API JavaScript library: [`near-api-js`](https://docs.near.org/docs/roles/developer/examples/near-api-js/introduction).
+
+  - All transactions require the following:
+      - `signerId` _(account ID of the transaction originator)_
+      - `signerPublicKey`
+      - `receiverId` _(account ID of the transaction recipient)_
+      - `nonceForPublicKey` _(each time a key is used the nonce value should be incremented by 1)_
+      - `actions` _( [[ click here ]](/docs/concepts/transaction#action) for supported arguments)_
+      - `blockHash` _(a current block hash (within 12hrs) to prove the transaction was recently created)_
+
+  See [Transaction Class](https://near.github.io/near-api-js/classes/_transaction_.transaction.html) for a more in depth outline.
+___
+
+# Create a "token transfer" transaction
+
+## Setup
+1) Clone the [transaction-examples](https://github.com/near-examples/transaction-examples) repository by running:
+
+```bash
+git clone https://github.com/near-examples/transaction-examples.git
+```
+
+2) Follow [setup instructions](https://github.com/near-examples/transaction-examples/blob/master/readme.MD#prerequisites)
+___
+
+## Imports
+In the first two lines of code in [`send-tokens.js`](https://github.com/near-examples/transaction-examples/blob/master/send-tokens.js#L1-L2), we:
+  1) Importing the NEAR API JavaScript library
+  2) Importing helper functions found in [`utils.js`](https://github.com/near-examples/transaction-examples/blob/master/utils.js)
+
+```js
+const nearAPI = require('near-api-js');
+const { setupNear, formatAmount } = require('./utils');
+```
+___
+
+## Setting up connection to NEAR
+[`utils.js`](https://github.com/near-examples/transaction-examples/blob/master/utils.js) has two key components to connect and perform transactions on NEAR.
+  1) Constructing a `keyStore` object using `.json` key pairs found in `/$HOME/near-credentials`.
+
+```js
+const keyStore = new nearAPI.keyStores.UnencryptedFileSystemKeyStore(
+  `${process.env.HOME}/.near-credentials`
+  );
+```
+  2) `setupNear()` constructs an object that will allow us to connect and interact with the NEAR blockchain. 
+  
+  - This function takes an an argument of `env` for the network environment. 
+  - `testnet` is setup by default, but you can modify this in [`send-tokens.js`](https://github.com/near-examples/transaction-examples/blob/master/send-tokens.js#L5).
+
+```js
+function setupNear(env){
+  const config = getConfig(env);
+  return new nearAPI.Near({
+    keyStore: keyStore,
+    networkId: config.networkId,
+    nodeUrl: config.nodeUrl,
+    walletUrl: config.walletUrl,
+    helperUrl: config.helperUrl,
+    explorerUrl: config.explorerUrl,
+  });  
+};
+```
+  **Note**: `setupNear()` retrieves network configuration settings from [`config.js`](https://github.com/near-examples/transaction-examples/blob/master/config.js) based on the argument passed. Also note that we use the `keyStore` created earlier to construct this new object.
+___
+
+## Formatting token amounts
+When sending NEAR tokens (Ⓝ) during a transaction, the amount needs to be converted into [Yocto](https://en.wikipedia.org/wiki/Yocto-) Ⓝ or (10^-24).
+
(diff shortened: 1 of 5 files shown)


docs/tutorials/create-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+---
+To construct & process transactions you will need our API JavaScript library: [`near-api-js`](https://docs.near.org/docs/roles/developer/examples/near-api-js/introduction).
+
+  - All transactions require the following:
+      - `signerId` _(account ID of the transaction originator)_
+      - `signerPublicKey`
+      - `receiverId` _(account ID of the transaction recipient)_
+      - `nonceForPublicKey` _(each time a key is used the nonce value should be incremented by 1)_
+      - `actions` _( [[ click here ]](/docs/concepts/transaction#action) for supported arguments)_
+      - `blockHash` _(a current block hash (within 12hrs) to prove the transaction was recently created)_
```
bowenwang1996, line 14:
```suggestion
      - `blockHash` _(a current block hash (within 24hrs) to prove the transaction was recently created)_
```

docs/tutorials/create-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+  1) Importing the NEAR API JavaScript library
+  2) Importing helper functions found in [`utils.js`](https://github.com/near-examples/transaction-examples/blob/master/utils.js)
+
+```js
+const nearAPI = require('near-api-js');
+const { setupNear, formatAmount } = require('./utils');
+```
+___
+
+## Setting up connection to NEAR
```
bowenwang1996, line 42: I would also add a section on using a keypair directly to construct a transaction (vs. going to keystore to find keys)

docs/tutorials/create-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+- For this example, we are using `Transfer`
+- This transfer action can be created using the [imported `nearAPI` object](/docs/tutorials/create-transactions#imports) and the [formatted Ⓝ amount](/docs/tutorials/create-transactions#formatting-token-amounts) created earlier.
+
+```js
+const actions = [nearAPI.transactions.transfer(amount)];
+```
+ [[ click here ]](https://github.com/near/near-api-js/blob/master/src/transaction.ts#L70-L72) to view source for `transfer()`.
+
+### 6 `blockHash`
+- Each transaction requires a current block hash (within 12hrs) to prove that the transaction was created recently.
```
bowenwang1996, line 156:
```suggestion
- Each transaction requires a current block hash (within 24hrs) to prove that the transaction was created recently.
```

docs/tutorials/create-transactions.md
thisisjoshford: fixed 👍

docs/tutorials/create-transactions.md
```diff
@@ -181,29 +181,45 @@ const tx = nearAPI.transactions.createTransaction(
...
-  );
+ const serializedTx = nearAPI.utils.serialize.serialize(
+    nearAPI.transactions.SCHEMA, 
+    transaction
+    );
 ```
 
+2) Hash the transaction using a `sha256` cryptographic hashing algorithm.
+```js
+const hashedTx = new Uint8Array(sha256.sha256.array(serializedTx));
```
frol, line 196: I think `hashedTx` is a bit misleading as the result is a hash of the serialized transaction blob (`serializedTxHash`)

docs/tutorials/create-transactions.md
```diff
@@ -181,29 +181,45 @@ const tx = nearAPI.transactions.createTransaction(
...
+```
+
+3) Create a unique signature with your `keyPair` and the hashed transaction.
+```js
+const signature = keyPair.sign(hashedTx);
+```
+
+4) Construct the signed transaction using `near-api-js` [SignedTransaction class](https://github.com/near/near-api-js/blob/master/src/transaction.ts#L112-L123).
```
frol, line 204: It is better to have links with a specific commit hash in them as `master` will get the file updated over time and the line references will not match.

frol (approved):
LGTM

docs/tutorials/create-transactions.md
thisisjoshford: Ahh, good call. Thanks @frol !

docs/tutorials/create-transactions.md
```diff
@@ -29,44 +29,44 @@ git clone https://github.com/near-examples/transaction-examples.git
...
 ```
-  2) `setupNear()` constructs an object that will allow us to connect and interact with the NEAR blockchain. 
-  
-  - This function takes an an argument of `env` for the network environment. 
-  - `testnet` is setup by default, but you can modify this in [`send-tokens.js`](https://github.com/near-examples/transaction-examples/blob/master/send-tokens.js#L5).
 
+___
+
+## Access Keys
+To sign a transaction as well as send NEAR Ⓝ, we will need a `FullAccess` key to the sender's account.
```
bowenwang1996, line 58:
```suggestion
To sign a transaction to send NEAR Ⓝ, we will need a `FullAccess` key to the sender's account. 
```

docs/tutorials/create-transactions.md
```diff
@@ -129,10 +129,10 @@ PublicKey = {
 ### 4 `nonceForPublicKey`
 - A unique number or `nonce` is required for each transaction performed with a public key.
```
bowenwang1996, line 130:
```suggestion
- A unique number or `nonce` is required for each transaction signed with an access key.
```

docs/tutorials/create-transactions.md
```diff
@@ -181,29 +181,45 @@ const tx = nearAPI.transactions.createTransaction(
...
+    transaction
+    );
 ```
 
+2) Hash the transaction using a `sha256` cryptographic hashing algorithm.
+```js
+const hashedTx = new Uint8Array(sha256.sha256.array(serializedTx));
+```
+
+3) Create a unique signature with your `keyPair` and the hashed transaction.
```
bowenwang1996, line 199:
```suggestion
3) Sign the hashed transaction. with your `keyPair`
```

docs/tutorials/create-transactions.md
```diff
@@ -181,29 +181,45 @@ const tx = nearAPI.transactions.createTransaction(
...
+      data: signature.signature 
+    })
+  });
+```
+
+
 ## Send transaction
 Final step is to encode and send the transaction.
-  - First we serialize transaction into [BORSH](https://borsh.io/), and store the result as `bytes`. _(required for all transactions)_
+  - First we serialize transaction into [BORSH](https://borsh.io/), and store the result as `serializedTx`. _(required for all transactions)_
```
bowenwang1996, line 218: I think we usually just write `Borsh` instead of `BORSH`

bowenwang1996 (approved):

PR #511, opened 2020-10-16
fix: update outdated github.com/nearprotocol links

chadoh:
And remove some unnecessary excluded URLs from the link checker

The second commit removes confusing duplication from how Travis checked links, which used a copy-and-pasted version of our `test-links.sh` script. See commit message for details.

### DIFF
--- a/website/test-links.sh
+++ b/website/test-links.sh
@@ -3,7 +3,8 @@
 set -e
 
 # build the project
-yarn build
+npm install
+npm run build
 
 # serve up files for testing to port 5000 (default)
 npx serve build/near-docs &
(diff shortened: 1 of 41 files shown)


website/test-links.sh
```diff
@@ -15,7 +15,7 @@ npx wait-on http://localhost:5000/
 set +e
 
 # crawl site and check all links
-npx broken-link-checker -gro --exclude "https://staking.dokia.cloud/staking/near/validators" --exclude "localhost:3030" --exclude "https://youtu.be" --exclude "http://near.ai/wbs" --exclude "https://www.youtube.com" --host-requests 2 http://localhost:5000 --exclude "https://rpc.testnet.near.org" --exclude "https://rpc.betanet.near.org" --exclude "https://crates.io/crates/near-sdk" --exclude "https://wasmbyexample.dev"
+npx broken-link-checker -gro --exclude "localhost:3030" --exclude "https://youtu.be" --exclude "http://near.ai/wbs" --exclude "https://www.youtube.com" --host-requests 2 http://localhost:5000 --exclude "https://rpc.testnet.near.org" --exclude "https://rpc.betanet.near.org" --exclude "https://crates.io/crates/near-sdk" --exclude "https://wasmbyexample.dev" --exclude "https://staking.dokia.cloud"
```
thisisjoshford, line 18: This needs to be added to `.travis.yml` as well :) I fixed this and two other problematic links with PR #513

chadoh:
Looks like a false positive on the test passing – if you check, you'll see that `rpc.near.org` shows up as broken right now, but Travis still thinks it passed, because the script exits with status 0.

chadoh:
ok, links updated and Travis updated to use `test-links.sh`. Duplicate Travis script removed. I think this is ready to merge. Do you see anything suspicious, @thisisjoshford?

thisisjoshford (approved):
LGTM! :)

PR #518, opened 2020-11-05
for NEAR CLI add section about package manager and nvm

mikedotexe:
As we have on the homepage, https://near.org, we should mention that for Linux/Mac users they should use a package manager.
I have not heard from folks that Windows needs a package manager. If we hear that feedback, we can modify the Windows section as well.

### DIFF
--- a/docs/development/near-cli.md
+++ b/docs/development/near-cli.md
@@ -71,8 +71,8 @@ _Click on a command for more information and examples._
 
 #### Mac and Linux
 
-1. Install `npm` [[ click here ]](https://www.npmjs.com/get-npm)
-2. Install `NodeJS` [[ click here ]](https://nodejs.org/en/download)
+1. Install `npm` It's recommended to use a package manager like `nvm` as sometimes there are issues using Ledger due to how OS X handles node packages related to USB devices. [[ click here ]](https://nodejs.org/en/download/package-manager/)
+2. Ensure you have installed Node version 12 or above.
 3. Install `near-cli` globally by running:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/development/near-cli.md
```diff
@@ -71,8 +71,8 @@ _Click on a command for more information and examples._
 
 #### Mac and Linux
 
-1. Install `npm` [[ click here ]](https://www.npmjs.com/get-npm)
-2. Install `NodeJS` [[ click here ]](https://nodejs.org/en/download)
+1. Install `npm` It's recommended to use a package manager like `nvm` as sometimes there are issues using Ledger due to how OS X handles node packages related to USB devices. [[ click here ]](https://nodejs.org/en/download/package-manager/)
```
thisisjoshford, line 74: did you mean to remove the `npm` installation link?  Was your intention to have the node package manager info for step 2?

mikedotexe: nvm will install node and npm, then I wanted to make sure they install node 12

docs/development/near-cli.md
```diff
@@ -71,8 +71,8 @@ _Click on a command for more information and examples._
...
+2. Ensure you have installed Node version 12 or above.
```
thisisjoshford, line 75: Maybe this?

```suggestion
2. Ensure you have installed `NodeJS` version 12 or above.
  **Note:**  For an optimal experience, we recommend using a package manager such as `nvm` as, without one, OS X has experienced issues using Ledger devices. This is due to the default way OS X handles node packages with USB devices. [[ click here ]](https://nodejs.org/en/download/package-manager/) to install.
```

docs/development/near-cli.md
thisisjoshford: ahh gotcha!

PR #534, opened 2020-11-13
Update examples.md

TrevorJTClarke:
I could be wrong here -- but seems like "message" was the intent here, not "hash" even tho thats what it is

### DIFF
--- a/docs/roles/developer/examples/near-api-js/examples.md
+++ b/docs/roles/developer/examples/near-api-js/examples.md
@@ -425,7 +425,7 @@ const bytes = transaction.encode();
 Sign transaction (offline with access to the key):
 
 ```js
-const hash = new Uint8Array(sha256.sha256.array(bytes));
+const message = new Uint8Array(sha256.sha256.array(bytes));
 const signature = await signer.signMessage(message, accountId, networkId);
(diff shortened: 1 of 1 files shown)


bowenwang1996 (approved):

docs/roles/developer/examples/near-api-js/examples.md
```diff
@@ -425,7 +425,7 @@ const bytes = transaction.encode();
 Sign transaction (offline with access to the key):
 
 ```js
-const hash = new Uint8Array(sha256.sha256.array(bytes));
+const message = new Uint8Array(sha256.sha256.array(bytes));
```
thisisjoshford, line 428:
```suggestion
const signedTxHash = new Uint8Array(sha256.sha256.array(bytes));
```

thisisjoshford:
> I could be wrong here -- but seems like "message" was the intent here, not "hash" even tho thats what it is

Thanks for bringing our attention to this @TrevorJTClarke !  This "offline transaction signing" example should really be redone in a clearer way and this doc, overall, should get revised and will be done this week :)

TrevorJTClarke:
sounds great! thanks for the update @thisisjoshford

PR #537, opened 2020-11-17
add evm-view/evm-call to NEAR CLI page

mikedotexe:
Will be merging and publishing NEAR CLI with these two new commands shortly.
Unfortunately these are a bit difficult to reproduce without an EVM setup. Happy to help if you'd like to try.


thisisjoshford (approved):
Awesome! Thanks for the update @mikedotexe !

PR #544, opened 2020-12-01
Fix nearcore contribution instructions

artob:
A quick fix for a couple of issues I noticed at https://docs.near.org/docs/contribution/nearcore:

- The `start_near.sh` script no longer exists, it appears to have been renamed to `start_localnet.py`.

- The `--verbose` option requires an argument value be provided.

Screen Shot 2020-12-01 at 18 37 02


frankbraun (approved):
I verified that these changes are necessary and correct.

thisisjoshford (approved):
awesome! thanks so much for the update :)

PR #545, opened 2020-12-01
Fix nearcore contribution typos

frankbraun:
* The http binary is part of the httpie package.

* rustmt -> rustfmt.


thisisjoshford (approved):
thanks for the fix!! :)

PR #548, opened 2020-12-07
Update examples.md

TrevorJTClarke:
I think this was a copy/paste error. Also added missing param needed for RPC call.

### DIFF
--- a/docs/roles/developer/examples/near-api-js/examples.md
+++ b/docs/roles/developer/examples/near-api-js/examples.md
@@ -284,8 +284,8 @@ The code snippet below is too short to be useful except as an illustration.  A b
 // best to fetch this transaction hash using a call to the live network
 // if this call fails it's because this transaction hash is no longer valid
 let txHash = "3L37n7LowxfpFEJcQvYmoQPo4NpgV2ZUmr4vXSBYhxPL"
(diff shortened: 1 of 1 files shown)


docs/roles/developer/examples/near-api-js/examples.md
```diff
@@ -284,8 +284,8 @@ The code snippet below is too short to be useful except as an illustration.  A b
 // best to fetch this transaction hash using a call to the live network
 // if this call fails it's because this transaction hash is no longer valid
 let txHash = "3L37n7LowxfpFEJcQvYmoQPo4NpgV2ZUmr4vXSBYhxPL"
-let decodedTxHash = nearApi.utils.serialize.base_decode("3L37n7LowxfpFEJcQvYmoQPo4NpgV2ZUmr4vXSBYhxPL")
-await near.connection.provider.txStatus(decodedTxHash, '')
+let decodedTxHash = near.utils.serialize.base_decode("3L37n7LowxfpFEJcQvYmoQPo4NpgV2ZUmr4vXSBYhxPL")
```
thisisjoshford, line 287:
```suggestion
let decodedTxHash = nearAPI.utils.serialize.base_decode("3L37n7LowxfpFEJcQvYmoQPo4NpgV2ZUmr4vXSBYhxPL")
```

`nearAPI` is the name we use for our examples when importing `near-api-js`

thisisjoshford (changes requested):
thanks for the update! this doc needs an overhaul, but the accountID should certainly be included here.

thisisjoshford (approved):

PR #555, opened 2020-12-12
update blockId vs finality for RPC endpoints

### LINKED ISSUE #549: [DOC] JSON RPC documentation is not explicit about the fact that block_id OR finality can be used
All the APIs which take `"block_id"` OR `"finality"` parameters can take either of those. The [`block` endpoint](https://docs.near.org/docs/api/rpc#block-details) is documented accurately, but `query`-based and "changes" APIs only list either "finality" or "block_id" as an option while in reality, you can pass either of those (both cannot be present at the same time since it does not make sense).

@thisisjoshford you have put the whole doc beautifully, may I ask you to come up with a human-readable representation of these available options that users may want to choose from?

thisisjoshford:
- added `block_id` and `finality` parameter descriptions
- clarified view and change methods that can take _either_ `block_id` _or_ `finality` with reference links

fixes #549

### DIFF
--- a/docs/api/rpc.md
+++ b/docs/api/rpc.md
@@ -69,7 +69,7 @@ http post https://rpc.testnet.near.org jsonrpc=2.0 id=dontcare method=network_in
 - method: `query`
 - params:
   - `request_type`: `view_access_key`
-  - [`finality`](/docs/api/rpc#using-finality-param) _OR_ [`block_Id`](/docs/api/rpc#using-block_id-param)
+  - [`finality`](/docs/api/rpc#using-finality-param) _OR_ [`block_id`](/docs/api/rpc#using-block_id-param)
   - `account_id`: _`"example.testnet"`_
   - `public_key`: _`"example.testnet's public key"`_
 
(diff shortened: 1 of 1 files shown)


docs/api/rpc.md
```diff
@@ -55,7 +69,7 @@ http post https://rpc.testnet.near.org jsonrpc=2.0 id=dontcare method=network_in
 - method: `query`
 - params:
   - `request_type`: `view_access_key`
-  - `finality`: `optimistic` or `final`
+  - [`finality`](/docs/api/rpc#using-finality-param) _OR_ [`block_Id`](/docs/api/rpc#using-block_id-param)
```
frol, line 72: Nit (apply to the other places as well):

```suggestion
  - [`finality`](/docs/api/rpc#using-finality-param) _OR_ [`block_id`](/docs/api/rpc#using-block_id-param)
```

frol (approved):
😍 

Beautiful! You are a wizard of documentation!

thisisjoshford:
> 😍
> 
> Beautiful! You are a wizard of documentation!

Thank you soooo much!  You are too kind! 🙂

PR #561, opened 2020-12-17
testnet gives 500 tokens, not 10

mikedotexe:
self-explanatory


thisisjoshford (approved):
Thanks Mike!  CI is failing due to something unrelated, so this is good to go. :)

PR #562, opened 2020-12-17
feat(api): Documented a new JSONRPC method to query contract code

### LINKED ISSUE #571: [DOC] Add view_code RPC endpoint to docs
There is a new endpoint `view_code` merged here:
https://github.com/near/nearcore/pull/3714

At the time of this writing it's live on betanet. I think it's fine to have the docs reflect an example that uses env vars for betanet.

http post https://rpc.betanet.near.org jsonrpc=2.0 id=dontcare method=query params:='{"request_type":"view_code","finality":"final","account_id":"mike.betanet"}'

You will get something like the attached output. Truncate the attachment 
[output.txt](https://github.com/near/docs/files/5798018/output.txt)
for the purpose of the example

frol:
This is not live yet, so we may want to merge it later.

/cc @ailisp 

*Resolves #571*


thisisjoshford (approved):
Awesome! Looks great and the endpoint is working on `betanet`... I take it we should wait until `mainnet` release before merging @frol ?

frol:
Well, it seems that we need some process here since it is hard to keep track of these. I think we can add some message that says that the feature is only available in canary (one can build and run such a node locally), but not available via our public RPC yet. WDYT? @thisisjoshford

frol:
It is available on testnet now.

@thisisjoshford may I ask you to resolve the conflict and merge?

PR #564, opened 2020-12-20
Fix [Storage Staking] Section Bug

think-in-universe:
### Bug Fixing

In the section of **Use a binary serialization format, rather than JSON** https://docs.near.org/docs/concepts/storage#use-a-binary-serialization-format-rather-than-json

When using a binary serialization format with `borsh` towards the array `[0, 1, 2, 3]`,  the last 4 bytes should be `\u0000\u0001\u0002\u0003` instead of `\u0000\u0001\u0002\u0004`

### DIFF
--- a/docs/concepts/storage.md
+++ b/docs/concepts/storage.md
@@ -115,7 +115,7 @@ Using borsh, this same array gets saved as 8 bytes:
 
 At first glance, saving 1 byte might not seem significant. But let's look closer.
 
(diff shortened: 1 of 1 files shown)


thisisjoshford (approved):
awesome, thanks for the catch!

think-in-universe:
@thisisjoshford  thanks for merge, Josh.

PR #570, opened 2021-01-11
add collection / data storage 

thisisjoshford:
- added collection / data storage types documentation to basics section
- created & linked example repositories ([`collection-examples-rs`](https://github.com/near-examples/collection-examples-rs) & [`collection-examples-as`](https://github.com/near-examples/collection-examples-as))


mikedotexe:
@thisisjoshford can you merge in master or rebase? the broken link checker doesn't seem to have the Twitter exclusion that' on master here: https://github.com/near/docs/blob/master/website/test-links.sh#L19

thisisjoshford:
> @thisisjoshford can you merge in master or rebase? the broken link checker doesn't seem to have the Twitter exclusion that's on master here: https://github.com/near/docs/blob/master/website/test-links.sh#L19

ahh was wondering why twitter was failing CI all of a sudden, had a note to whitelisted it but must have already been done earlier :) thx

mikedotexe:
Here:
https://docs.rs/near-sdk/2.0.0/near_sdk/collections/struct.Vector.html#method.clear
We see that the Rust Vector does have a `clear` method. We'll need to modify the "Clear all values" column shown here:
https://docs-pr-570.onrender.com/docs/concepts/data-storage

mikedotexe:
Rust Vector preserves order as far as I know, too.

mikedotexe:
@willemneal could you do a quick review on what the highlighted rows should look like in this chart?

Screen Shot 2021-01-11 at 3 57 17 PM

bowenwang1996:
I don't think self-balancing should be one of the columns since it doesn't even make sense when the data structure is not a tree. It is better to think in terms of complexity of operations such as insertion, deletion, etc.

thisisjoshford:
> I don't think self-balancing should be one of the columns since it doesn't even make sense when the data structure is not a tree. It is better to think in terms of complexity of operations such as insertion, deletion, etc.

Sounds good, I'll remove that column. Is there another column that you would suggest with regards to operation complexity?

docs/concepts/data-collections.md
```diff
@@ -0,0 +1,328 @@
...
+> All data stored on the NEAR blockchain is done in key / value pairs. There are several collection methods in the SDKs we've created that will help you store your data on chain.
+>
+> - [`near-sdk-as`](https://github.com/near/near-sdk-as) for [AssemblyScript](https://www.assemblyscript.org/) smart contracts
+> - [`near-sdk-rs`](https://github.com/near/near-sdk-rs) for [Rust](https://www.rust-lang.org/) smart contracts
+
+---
+
+## AssemblyScript Collection Types
+
+| Type                                                                           | Iterable | Clear all values | Preserves order |
```
willemneal, line 18: Should mean order of keys.  Which is insertion order in the case of the unordered map and sorted order in terms of the trees.

docs/concepts/data-collections.md
```diff
@@ -0,0 +1,328 @@
...
+
+**Note:** The Map doesn't store keys, so if you need to retrieve them, include keys in the values.
+
+[ [SDK source](https://github.com/near/near-sdk-as/blob/master/sdk-core/assembly/collections/persistentMap.ts) ]
+
+---
+
+### `PersistentUnorderedMap`
+
+> Implements an unordered map built on top of the [`Storage`](https://github.com/near/near-sdk-as/blob/master/sdk-core/assembly/storage.ts) class.
```
willemneal, line 97:
```suggestion
> Implements an unordered map built on top of the [`PersistentMap`](https://github.com/near/near-sdk-as/blob/master/sdk-core/assembly/collections/persistentMap.ts) class and a [`PersistentVector`](https://github.com/near/near-sdk-as/blob/master/sdk-core/assembly/collections/persistentVector.ts), which stores the keys of the map.  These keys are initially in the order they are inserted, however, a deleted key is swapped with the last key.
```

willemneal (commented):
I agree with @bowenwang1996, the comparisons in the table should be more of a list of API's and algorithmic costs.

mikedotexe:
In kind of a funsie thought, I'd love to see something other than the character "x" or "X" like:
✘
```
<p>I will display &#10008;</p>
<p>I will display &#x2718;</p>
```
See list of HTML entities here:
https://www.w3schools.com/charsets/ref_utf_dingbats.asp

docs/concepts/data-collections.md
thisisjoshford: updated :)

PR #584, opened 2021-01-21
revise / enhance networks documentation

### LINKED ISSUE #578: [BUG] Betanet description not accurate
>BetaNet is intended primarily for validators who wish to test their infrastructure.

That's not true since October. Remove that and perhaps review this page again for errors.
https://docs.near.org/docs/roles/developer/networks#betanet

thisisjoshford:
Fixes #578

### DIFF
--- a/docs/roles/developer/networks.md
+++ b/docs/roles/developer/networks.md
@@ -33,7 +33,7 @@ All accounts on `mainnet` have a `.near` top level [account name](/docs/concepts
 
 All accounts on `testnet` have a `.testnet` top level [account name](/docs/concepts/account) and use the following tools:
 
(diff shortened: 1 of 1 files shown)


bowenwang1996 (commented):
I would not say that "all accounts on testnet have a .testnet suffix" since there are still other accounts. The .testnet and .betanet suffixes only apply to accounts created from wallet

thisisjoshford:
> I would not say that "all accounts on testnet have a .testnet suffix" since there are still other accounts. The .testnet and .betanet suffixes only apply to accounts created from wallet

Ahh good point! Thx @bowenwang1996 :)

PR #620, opened 2021-02-19
Fix transfer token command.

veneris:


thisisjoshford (approved):
Awesome! Nice catch :) Thanks for the PR!

PR #627, opened 2021-03-05
Update near-cli.md

blake41:
Is this the best way to ask a question like this?


blake41:
But then there's no chance you will update the documentation with an explanation with how it works...

PR #635, opened 2021-03-15
Fix: Make a Connection (Browser) invalid config

tahpot:


thisisjoshford (approved):
awesome! thanks for the PR! :)

PR #640, opened 2021-03-16
add language tabs to rpc

### LINKED ISSUE #639: [DOC] Update RPC docs to include JavaScript snippets
Add `near-api-js` documentation to the [RPC docs](https://docs.near.org/docs/develop/front-end/rpc). This can be done using language tabs in the documentation:

Screenshot from 2021-03-15 22-41-41

Screenshot from 2021-03-15 22-41-48

mehtaphysical:
Fixes #639.

Not sure if we want this. But here is an example. Its not 100% accurate right now. Some of the `near-api-js` provider methods are missing. See https://github.com/near/near-api-js/issues/532.


thisisjoshford (approved):
I absolutely love this approach! Thanks so much, @mehtaphysical :) 

I'm going to wait on merging this PR until I complete near/near-api-js/issues/532 which will ensure all RPC endpoints are available in `near-api-js`.

PR #641, opened 2021-03-16
todo CRUD tutorial

mehtaphysical:


docs/tutorials/apps/todos-crud-app.md
```diff
@@ -0,0 +1,858 @@
...
+title: Building a CRUD App
+sidebar_label: Building a CRUD App
+---
+
+## Overview
+
+In our application we'll have two distinct layers:
+
+1. smart contract (in web2 we may refer to this as server-side or back-end)
+1. web app (in web2 we may refer to this as client-side or front-end)
```
thisisjoshford, line 12:
```suggestion
2. web app (in web2 we may refer to this as client-side or front-end)
```

BenKurrek:
- The first sentence of 'Building the App' might be better worded as: We're building a Create-Read-Update-Delete (CRUD) application and we'll need to add smart contract methods for each of these operations"
- perhaps a high-level explanation of what the CRUD API will be used for at the beginning would be nice so maybe mentioning the todo aspect.
- In the create/model section, you had previously talked about using a create method to make a task. You then talk about storing the todos and as a user, I got a little confused as to what the difference between the create and the store was. Maybe this goes back to the point i made above about the high level explanation of what the API will be used for? 
- OK so now I understand what is going on - u created an example test case using the create method but the create method hasn't actually been implemented yet and so that's why I got confused - perhaps something to think about.
- Once i understood the above, everything else was smooth sailing. 

TLDR - I got a little lost in the sauce due to formatting but once I understood, it was smooooth as butter :) great job Ryan!

thisisjoshford:
Started going through this tutorial and I see that the AseemblyScript steps are outdated @bucanero. As I was making updates, it occurred to me that perhaps we should re-do this tutorial in Rust as we are currently encouraging developers to use `near-sdk-rs` over `near-sdk-as` until AS becomes a little more mature. Thoughts?

thisisjoshford (approved):

PR #646, opened 2021-03-22
update storage staking price to 1E19

### LINKED ISSUE #644: [BUG] Storage Staking Cost changed
**Describe the bug** 🐞
The storage stacking costs have changed from `1E20` to `1E19` as of [1.18.0](https://github.com/near/nearcore/releases/tag/1.18.0)

We should change the cost in the docs, but what should the link point to? It can't point to the genesis config because that will remain `1E20`. Should the link be removed? Should the link point to the [governance discussion](https://gov.near.org/t/storage-staking-price/399)?

**Expected behavior**
change storage stacking cost from `1E20` to `1E19`

**Location** 🎯
https://docs.near.org/docs/concepts/storage-staking

mehtaphysical:
Fixes #644


thisisjoshford (approved):
thanks!

PR #658, opened 2021-04-14
fixed: Case error

little-buddy:
line#64
const { keyStores, keyPair } = nearAPI
=>
const { keyStores, KeyPair } = nearAPI;


thisisjoshford (approved):
Nice catch! thanks @little-buddy ! 🙏

PR #659, opened 2021-04-15
Fix load contract parameters bug

think-in-universe:
### Description

The **nearAPI.Contract** constructor accept the parameters as below. 

```ts
const Contract: new (account: nearAPI.Account, contractId: string, options: {
    viewMethods: string[];
    changeMethods: string[];
}) => nearAPI.Contract
```

The 1st parameter should be an account object, and the 2nd `config` is not necessary.


thisisjoshford (approved):
nice catch! thanks for the PR! 🙏

PR #663, opened 2021-04-20
Corrected descriptions of network endpoints of JSON RPC

frol:
Note: I moved the "Node Status" section to be the first since it is the most useful one


thisisjoshford (approved):
Awesome! Thanks @frol 🙏

frol:
@thisisjoshford Wow, thanks for the review and merge! That was blazing fast! :zap:

Issue #660, opened 2021-04-15
[BUG] The default allowance of adding key with near-cli is not 0.25


think-in-universe:
**Describe the bug** 🐞

As reported by a developer, the description of the allowance argument in https://docs.near.org/docs/tools/near-cli#2-add-a-function-access-key  is incorrect or out of date.

`--allowance is the amount of Ⓝ the key is allowed to spend on gas fees only and if omitted, 0.25 is the default.`

The allowance is N/A rather than 0.25, if he adds key via near-cli but didn't specify the allowance, e.g.

command: `near add-key example-acct.testnet GkMNfc92fwM1AmwH1MTjF4b7UZuceamsq96XPkHsQ9vi --contract-id example-contract.testnet`

The allowance is N/A in web wallet: 

![image](https://user-images.githubusercontent.com/46699230/114879688-01c35a80-9e34-11eb-9db8-c6b23469f76c.png)

By looking at the source code, the **near-cli** (https://github.com/near/near-cli/blob/aa7642d414c8a8ee8737e46e9d96c2f5dd002122/commands/add-key.js#L34 ) and **near-api-js** (https://github.com/near/near-api-js/blob/71920e064febe9087c99bcc6a2c17efbfc36b1a5/src/account.ts#L316) didn't set the allowance by default. 

**Expected behavior**

We can fix either the doc by removing the description about the default allowance, or adding the default 0.25 $NEAR allowance in **near-cli** or **near-api-js** .

**Screenshots**
already attached above

**Location** 🎯
https://docs.near.org/docs/tools/near-cli#2-add-a-function-access-key

thisisjoshford:
Just tested this and you're correct @think-in-universe! It seems wallet creates a default of 0.25 N but `near-cli` & `near-api-js` does not. I'm going to update the docs to correct this and will bring this up in the next tooling meeting to see if we should create a default value for this arg.

thisisjoshford:
> Just tested this and you're correct @think-in-universe! It seems wallet creates a default of 0.25 N but `near-cli` & `near-api-js` does not. I'm going to update the docs to correct this and will bring this up in the next tooling meeting to see if we should create a default value for this arg.

thinking about it more @think-in-universe ... there _would_ be use cases where you would want to create a key without any allowance if the key is intended to be used for view call methods only. I updated the PR to include mention of this but will still bring it up during Dev Platform sync tomorrow.

think-in-universe:
@thisisjoshford thanks for the update, Josh.  

[...]

PR #690, opened 2021-05-13
Fix description of allowance in function call key

think-in-universe:
### Description

Find this issue when reviewing the document: https://medium.com/nearprotocol/an-introduction-to-near-accounts-f96cb84ef091

For `allowance` of function call key:

1. 0.25 default is not right as we discussed in https://github.com/near/docs/issues/660 I changed it to "required" but seems not that accurate as well, since the default value should be "null" which allows calling view methods only. 
2. the description of `allowance` seems not accurate for **amount of Ⓝ loaded onto the key**

### DIFF
--- a/docs/concepts/account.md
+++ b/docs/concepts/account.md
@@ -109,7 +109,7 @@ A `FunctionCall` key is unique as it only has permission to call a smart contrac
 1) `allowance` - the amount of Ⓝ the key is allowed to spend on gas fees _(optional - default: `null`)_
 2) `receiver_id` - contract the key is allowed to call methods on _(required)_
 3) `method_names` - contract methods the key is allowed to call _(optional)_
-
+**Note:** If `allowance` is omitted the default will be `null` and key will only be allowed to call view methods. Allowance can not be added after key is created.
 > **Note:** If no specific method names are specified, all methods may be called.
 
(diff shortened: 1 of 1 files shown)


docs/concepts/account.md
```diff
@@ -106,7 +106,7 @@ See our [action specifications](https://nomicon.io/RuntimeSpec/Actions.html) sec
 
 A `FunctionCall` key is unique as it only has permission to call a smart contract's method(s) that _do not_ attach Ⓝ as a deposit (i.e. payable functions). These keys have the following three attributes:
 
-1) `allowance` - amount of Ⓝ loaded onto the key to pay for gas fees  _(0.25 default)_
+1) `allowance` - the amount of Ⓝ the key is allowed to spend on gas fees _(required)_
```
thisisjoshford, line 109:
```suggestion
1) `allowance` - the amount of Ⓝ the key is allowed to spend on gas fees _(optional - default: `null`)_
```

As you stated, this field isn't actually required. We should note the default value of `null` for now until this is updated with a default value.

docs/concepts/account.md
```diff
@@ -106,7 +106,7 @@ See our [action specifications](https://nomicon.io/RuntimeSpec/Actions.html) sec
...
 2) `receiver_id` - contract the key is allowed to call methods on _(required)_
 3) `method_names` - contract methods the key is allowed to call _(optional)_
```
thisisjoshford, line 112: Maybe we can add a note here:
```suggestion
**Note:** If `allowance` is omitted the default will be `null` and key will only be allowed to call view methods. Allowance can not be added after key is created.
```

thisisjoshford (changes requested):

thisisjoshford (approved):
committed two suggestions to merge this PR @think-in-universe .

PR #695, opened 2021-05-15
docs(API js): typo: ketStores instead of keyStores

rodmoreno:


thisisjoshford (approved):
nice catch! thanks for the PR @rodmoreno !!! 🙏🏼

PR #706, opened 2021-05-21
Docs for how to start a sandbox node and run e2e test on it

ailisp:
Add a doc on how to start a sandbox node and how to run e2e test on it. It's going to supersede end-to-end test sections once our tooling is fully ready that support  run test on both sandbox and testnet with one cli command. 

Update:
added a subsection of how to use sandbox-only feature (right now there's only one feature: patch_state)


matklad:
cc @DiscRiskandBisque (am I correct that you do want to be cc'ed for such design/UX aspects of this work?)

DiscRiskandBisque:
> cc @DiscRiskandBisque (am I correct that you do want to be cc'ed for such design/UX aspects of this work?)

Yes and thank you! Apologies for the delay, I need to set up better GitHub notifications.

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+
+test();
+```
+
+The test itself is very straightfoward. It does:
+
+1. Create testing accounts and deploy contract.
+2. Alice sign the transaction to set status, and get status to see set status works.
+3. Get Bob's status and that should be null because Bob has not set status.
+4. Bob set status and get status, should show Bob's status changed and not affect Alice status.
```
thisisjoshford, line 200:
```suggestion
1. Creates testing accounts and deploys a contract.
2. Performs a `set_status` transaction signed by Alice and then calls `get_status` to confirm `set_status` worked.
3. Gets Bob's status and which should be `null` as Bob has not yet set status.
4. Performs a `set_status` transaction signed by Bob and then calls `get_status` to show Bob's changed status and should not affect Alice's status.
```

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+
+> Most of the code above is boilerplate setup code to setup NEAR API, key pairs, testing accounts and deploy the contract. We're working on a near-cli `near test` command to do these setup code so you can focus on writing only `test()` for this kind of test.
```
thisisjoshford, line 202:
```suggestion
> Most of the code above is boilerplate code to setup NEAR API, key pairs, testing accounts, and deploy the contract. We're working on a near-cli `near test` command to do this setup code so you can focus on writing only `test()` for this.
```

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+
+## Sandbox-only Features for Testing
+
+If you only use above test script that only uses standard NEAR rpcs, your tests can also be executed on a testnet node. Just replace the network ID, node url, key path and account names in above script and rerun. There's also some additional Sandbox-only features that make certain tests easier. We'll review some examples in this section.
```
thisisjoshford, line 206:
```suggestion
If you only use the above test script that just uses standard NEAR RPCs your tests can also be executed on a `testnet` node. Simply replace the network ID, node url, key path, and account names in the above script and rerun the tests. There's also some additional "Sandbox only" features that make certain tests easier. We'll review some examples of those in the following section.
```

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+
+### Patch State on the Fly
+
+You can add or modify any contract state, contract code, account or access key during the test with `sandbox_patch_state` RPC.
+
+For arbitrary mutation on contract state, you cannot do it via transactions since transaction can only include contract calls that mutate state in contract programmed way. For example, for a NFT contract, you can do some operation with your owned NFTs but you cannot manipulate other people owned NFTs because smart contract has coded with checks to reject that. This is the expected behavior of the NFT contract. However you may want to change other people's NFT for test setup. This is called "arbitary mutation on contract state" and can be done by `sandbox_patch_state` RPC or stop the node, dump state as genesis, edit genesis and restart the node. The later approach is more complicated to do and also cannot be done without restart the node.
```
thisisjoshford, line 212:
```suggestion
For arbitrary mutation on contract state you cannot perform this with transactions as transactions can only include contract calls that mutate state in a contract programmed way. For example with an NFT contract, you can perform some operation with NFTs you have ownership of but you cannot manipulate NFTs that are owned by other accounts as the smart contract is coded with checks to reject that. This is the expected behavior of the NFT contract. However you may want to change another person's NFT for a test setup. This is called "arbitrary mutation on contract state" and can be done by the `sandbox_patch_state` RPC. Alternatively you can stop the node, dump state at genesis, edit genesis, and restart the node. The later approach is more complicated to do and also cannot be performed without restarting the node.
```

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+
+For patch contract code, account or access key, you can add them with normal deploy contract, create account or add key actions in transaction, but that's also limited to your account or sub-account. `sandbox_patch_state` RPC does not have this restriction.
```
thisisjoshford, line 214:
```suggestion
For patch contract code, account, or access keys you can add them with a normal deploy contract, create account, or add key actions in a transaction but that's also limited to your account or sub-account. `sandbox_patch_state` RPC does not have this restriction.
```

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+
+Let's see an example of how patch_state would help in a test. Assume you want to mock the real state of a mainnet, where `alice.near` has set a status, and you want to retrieve that message. Above script doesn't work out of box, because your master account is `test.near` and you can only
+create account of `alice.test.near`, not `alice.near`. Patch state can solve this problem.
```
thisisjoshford, line 217:
```suggestion
Let's explore an example of how `patch_state` would help in a test. Assume you want to mock the real state of `mainnet` where `alice.near` has set a status and you want to retrieve that status message. The above script doesn't work out of box because your master account is `test.near` and you can only create an account of `alice.test.near` not `alice.near`. Patch state can solve this problem.
```

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+
+1. Fetch current state from sandbox node. You can also do this with `sendJsonRpc` of `near-api-js`, or with any http client from command line:
```
thisisjoshford, line 219:
```suggestion
Here is a guide on running `patch_state`:

1. Fetch the current state from the sandbox node: _(You can also do this with `sendJsonRpc` of `near-api-js` or with any http client from command line)_
```

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+
+```bash
+$ curl http://localhost:3030 -H 'content-type: application/json' -d '{"jsonrpc": "2.0", "id":1, "method":"query", "params":{"request_type":"view_state","finality":"final", "account_id":"status-message.test.near","prefix_base64":""}}'
+
+{"jsonrpc":"2.0","result":{"values":[{"key":"U1RBVEU=","value":"AgAAAA8AAABhbGljZS50ZXN0Lm5lYXIFAAAAaGVsbG8NAAAAYm9iLnRlc3QubmVhcgUAAAB3b3JsZA==","proof":[]}],"proof":[],"block_height":24229,"block_hash":"XeCMK1jLNCu2UbkAKk1LLXEQVqvUASLoxSEz1YVBfGH"},"id":1}
+```
+
+You can see the contract only has one key-value pair in state, looks like base64 encoded. Let's figure out what it is.
```
thisisjoshford, line 227:
```suggestion
You can see the contract only has one key-value pair in state which looks like base64 encoded. Let's figure out what it is.
```

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+        "account_id": "status-message.test.near",
+        "data_key": "U1RBVEU=",
+        "value": "AwAAAA8AAABhbGljZS50ZXN0Lm5lYXIFAAAAaGVsbG8NAAAAYm9iLnRlc3QubmVhcgUAAAB3b3JsZAoAAABhbGljZS5uZWFyCwAAAGhlbGxvIHdvcmxk"
+      }
+    }
+  ]
+}}'
+```
+
+7. Now we can back to the test file and rerun the test with new state.
```
thisisjoshford, line 326:
```suggestion
7. Now we can go back to the test file and rerun the test with the new state.
```

thisisjoshford (changes requested):
Left some grammar edits, suggestions, and comments :)

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+
+```text
+set_status(message: string)
+get_status(account_id: string) -> string or null
+```
+
+`set_status` method stores message string under the sender's account name key on chain. And get_status retrieve message of given account name. If it were set by `set_status` before, it returns that string, otherwise it returns null.
+
+1. Start a near-sandbox node. If you have run a sandbox node with some tests before, delete `/tmp/near-sandbox` before start near-sandbox.
+2. Assume you clone the repo with status-message and source code stays in `status-message/`. The compiled contract lives in `status-message/res`. Let's do some preparation for the test:
```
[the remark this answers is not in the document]
ailisp: Yep good point, would require user not only just download the wasm file but clone the repo at the beginning

docs/develop/contracts/sandbox.md
```diff
@@ -0,0 +1,370 @@
...
+    account_id: "bob.test.near",
+  });
+  assert.equal(bob_message, "world");
+  alice_message = await aliceUseContract.get_status({
+    account_id: "alice.test.near",
+  });
+  assert.equal(alice_message, "hello");
+}
+
+test();
```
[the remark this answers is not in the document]
ailisp: Yeah this file seems too lengthy here, put in near-sdk-rs is better, The eventual ideal place is have this implemented as some utility functions in near-api-js, to make it as short as only contains `test()` above

docs/develop/contracts/sandbox.md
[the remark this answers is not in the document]
ailisp: Rewrite it to ask user clone near-sdk-rs at beginning

docs/develop/contracts/sandbox.md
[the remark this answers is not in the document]
ailisp: Make the test to near/near-sdk-rs#433, once merged, will remove here

matklad:
Finally got to reviewing this, sorry that it took me so long. 

This looks good to me overall. 

However, I am not sure what are concrete next steps here. The current state is very a proof-of-concept. It works, but the UX is pretty horrible in itself, and we don't give any explicit guidance for contract authors about the interactions between this new style of testing, and existing simulatior/unit-test. 

As such, I don't think we should be publishing this as a user-visible documentation, as it seems that that'll just create confusion over *currently* supported testing approaches. 

It seems like this is the point where the work should be taken over by *someone* from the devx team, but I don't know who that would be, cc @mikedotexe.

chefsale (approved):

ailisp:
@matklad 's point makes sense too. My opinion is present both manual, low level way and a high level, automatic way to the user, with advanced user might be interested to learn when high level way doesn't work, e.g. when the automatic way has a bug or a special case that automatic way doesn't cover. Right now we don't have truffle equivalent so this doc only has the low level way. Once we have, then the high level way should be add to the first part as of this document so user can see the short way first.

The other PR that documents repro testnet transaction with sandbox have applied this approach: https://github.com/near/docs/pull/724/files. First sections give the instructions on how to use a automatic script to fetch state for reproduce. The later section describes which parts of blockchain information will affect the transaction result and may help curious user to understand the low level and troubleshoot if it there's still difference in transaction repro.

ailisp:
Seems latest master d894e828737ff06d74a7dcf3403a0d2019de31c8 broke CI / test-links, let me see if i can make a fix to master

ailisp:
This one is now blocked by near/near-sdk-rs#433

thisisjoshford (approved):

ailisp:
@thisisjoshford would need your or Mike's approval on near-examples/rust-status-message#64, thanks

mikedotexe (approved):

PR #709, opened 2021-05-24
Minor typo and visual improvements

ktecho:
Small language improvement and give another CR so explanation of each point prints in a new line


thisisjoshford (approved):
awesome! thanks for the contribution @ktecho ! :)

PR #722, opened 2021-05-27
Add fungible token information to exchange integration page

mikedotexe:
We need to provide a way to allow exchanges to query information about fungible tokens. Particularly balances and when transfers happen. We want to be as specific as possible and include RPC calls instead of abstractions of our library or tools.

- [ ] Document simple transfer using the recommendations [in this comment](https://github.com/near/docs/pull/722#discussion_r640723627).
- [ ] Document transfer and call using the recommendations [in this comment](https://github.com/near/docs/pull/722#discussion_r640744353).


mikedotexe:
I've provided a little skeleton of the sections here and fleshed out the first one.
Useful links:
- https://nomicon.io/Standards/FungibleToken/Core.html
- https://github.com/near-examples/FT

volovyk-s (approved):

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,92 @@ and we can find its outcome in `receipts_outcome`:
...
+
+`dtool a2h '[34,49,48,48,48,48,48,48,34]' | dtool h2s`
+
+Returns:
+
+`"1000000"`
+
+Note that the fungible token balance of the account `mike.testnet` is `1000000` wrapped in double-quotes. This is because of an issue with JSON serialization. Amounts given in arguments and results must be serialized as Base-10 strings, e.g. "100". This is done to avoid JSON limitation of max integer value of 2**53, which can certainly happen with fungible tokens.
+
+### Simple transfer
```
mikedotexe, line 849: For this section, @volovyk-s please follow the instructions (I think you actually wrote this part) from the FT example. https://github.com/near-examples/FT
Create a fungible token, then send to another user. For instance, this will add storage for `mike.testnet` to the FT contract deployed to `ft.demo.testnet`:

```bash
near call ft.demo.testnet storage_deposit '{"account_id": "mike.testnet", "registration_only": true}' --accountId ft.demo.testnet --amount .025
near call ft.demo.testnet ft_transfer '{"receiver_id": "mike.testnet", "amount": "1000000"}' --accountId ft.demo.testnet --amount 0.000000000000000000000001
```

Now `mike.testnet` has `1000000` tokens.
After that last NEAR CLI command of `ft_transfer`, open up explorer (maybe this will help understand visually) and also get the info of the transaction using NEAR CLI's `tx-status` command. The exchanges will likely be doing lower-level RPC calls that return values like what we get from the `tx-status` command.

Look through the big list of the transaction status and determine how exchanges can tell that this was successful. (I think it's something like Status » SuccessValue or something)

Next, run the same kind of command, except make it error. Maybe send more tokens than are allowed, put in the wrong parameters, or send to an account that doesn't exist or isn't registered. 
Do the same thing where you'll inspect the transaction and look for ways exchanges can tell that this failed. (I think it's something like Status » Failed)

mikedotexe: Probably makes sense to have subheaders for Successful/Failed transfers, like how I have that below with "Transfer and call"

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,92 @@ and we can find its outcome in `receipts_outcome`:
...
+
+### Transfer and call
```
mikedotexe, line 851: Similar to the previous section, we'll need to deploy a fungible token to testnet or use the same one from previously. This time, however, we'll want to add another contract. So make a subaccount or whatever and deploy a contract to it. You will be writing this contract. It will be similar to this:
https://github.com/near/near-sdk-rs/blob/master/examples/fungible-token/test-contract-defi/src/lib.rs#L46-L79

I highlighted the key method here which is important. You can simply copy this whole example, remove everything else except this method, and have it return "1" (spoiler: `PromiseOrValue::Value(U128::from(1))`)

Then use NEAR CLI to use the "transfer and call" method on the fungible token (deployed earlier) to send 10 fungible tokens where the receiving contract is this new contract.

Again look at the transaction in Explorer and using NEAR CLI to determine that 9 fungible tokens were transferred. 
(With NEAR CLI you intended to send 10 tokens, then the contract said, "no thanks, I only need 9 and so I'm returning you 1 of them" and in the end only 9 are transferred.)

Show how the results of the promises in this chain of events can be looked at and determined that 9 are sent. You may have to do some base64 decoding (or maybe it's base58) on some stuff you see from `near tx-status …`

Reference: https://nomicon.io/Standards/FungibleToken/Core.html?search=53#reference-level-explanation

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,92 @@ and we can find its outcome in `receipts_outcome`:
...
-## Block and Finality
+## Fungible tokens
+
+Please see the [spec for the fungible token standard](https://nomicon.io/Standards/FungibleToken/README.html) and an [example implementation](https://github.com/near-examples/FT) for reference details.
+
+One notable aspect of the standard is that method names are prefixed with `ft_`. This will be a helpful convention when querying for transactions related to fungible tokens.
+
+### Get balance
+
+Using the abstraction of the NEAR CLI tool, we can check the balance of a user's account with this command:
```
thisisjoshford, line 782:
```suggestion
Using the abstraction of the [NEAR CLI](/docs/tools/near-cli) tool, we can check the balance of a user's account with [`near view`](/docs/tools/near-cli#near-view):
```

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,92 @@ and we can find its outcome in `receipts_outcome`:
...
+`near view ft.demo.testnet ft_balance_of '{"account_id": "mike.testnet"}'`
+
+Returns:
+
+```
+View call: ft.demo.testnet.ft_balance_of({"account_id": "mike.testnet"})
+'1000000'
+```
+
+Using HTTPie:
```
thisisjoshford, line 793:
```suggestion
Alternatively, you can [call a contract function](/docs/api/rpc#call-a-contract-function) using the `query` RPC endpoint. Below is an example using Httpie:
```

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,92 @@ and we can find its outcome in `receipts_outcome`:
...
+            48,
+            48,
+            48,
+            34
+        ]
+    }
+}
+```
+
+As mentioned earlier, the `result` is an array of bytes. There are various ways to convert bytes into a more human-readable form, such as the [dtool CLI](https://github.com/guoxbin/dtool#installation).
```
thisisjoshford, line 839:
```suggestion
As mentioned earlier, the `result` is an array of bytes. There are various ways to convert bytes into a more human-readable form such as the [dtool CLI](https://github.com/guoxbin/dtool#installation).
```

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,92 @@ and we can find its outcome in `receipts_outcome`:
...
+
```
thisisjoshford, line 840:
```suggestion

Example:

```

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,92 @@ and we can find its outcome in `receipts_outcome`:
...
+`dtool a2h '[34,49,48,48,48,48,48,48,34]' | dtool h2s`
+
+Returns:
+
+`"1000000"`
+
+Note that the fungible token balance of the account `mike.testnet` is `1000000` wrapped in double-quotes. This is because of an issue with JSON serialization. Amounts given in arguments and results must be serialized as Base-10 strings, e.g. "100". This is done to avoid JSON limitation of max integer value of 2**53, which can certainly happen with fungible tokens.
```
thisisjoshford, line 847:
```suggestion
**Note:** The fungible token balance of the account `mike.testnet` is `1000000` wrapped in double-quotes. This is because of an issue with JSON serialization. Amounts given in arguments and results must be serialized as Base-10 strings, e.g. "100". This is done to avoid JSON limitation of max integer value of 2**53, which can certainly happen with fungible tokens.
```

docs/roles/integrator/exchange-integration.md
```diff
@@ -810,9 +891,9 @@ The config should contain the following fields, currently NEAR testnet and mainn
 
 In the future there will be the possibility to track different or multiple shards.
 
-- Once the config has been changed you can restart the node and the node will start syncing new archival data, in the case where you want the full archival history you can just delete the data dir and start the node from scratch syncing full history or use one of the latest backups containing the data directory snapshot which can be copied under the near home dir (default: ~/.near/data).
+- Once the config has been changed you can restart the node it will start syncing new archival data, in the case where you want the full archival history you can just delete the data dir and start the node from scratch syncing full history or use one of the latest backups containing the data directory snapshot which can be copied under the near home dir (default: ~/.near/data).
```
thisisjoshford, line 894:
```suggestion
- Once the config has been changed you can restart the node it will start syncing new archival data. If you want the full archival history you can simply delete the data dir and start the node from scratch syncing full history. To speed up this process you can use one of the latest backups containing the data directory snapshot which can be copied under the near home dir (default: ~/.near/data).
```

thisisjoshford (changes requested):
Added some details / suggestions :)

docs/roles/integrator/exchange-integration.md
```diff
@@ -4,20 +4,19 @@ title: Exchange Integration
...
-### [Constructing Transactions](/docs/tutorials/create-transactions)
+Balance changes on accounts can be tracked by using our [changes endpoint](/docs/api/rpc#view-account-changes). You can test this out by sending tokens to an account using [NEAR CLI](/docs/tools/near-cli).
 
-## Balance Changes
-
-Balance changes on accounts can be tracked by using our [changes endpoint](/docs/api/rpc#changes). You can test this out by sending tokens to an account using [`near-cli](/docs/tools/near-cli).
-
-- First, make sure you have keys to your account locally.
-- Then send tokens using the following format. (The number at the end represents the amount you are sending)
+- First, make sure you have keys to your account locally. The typical workflow is to set up an account at https://wallet.testnet.near.org, then run the NEAR CLI command `login`. (Example: `near login`)
```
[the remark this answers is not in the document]
mikedotexe: Weird that I can't edit this, but we're missing the closing three-backticks, yes?

docs/roles/integrator/exchange-integration.md
```diff
@@ -64,18 +63,17 @@ Your response should look like this:
...
 
-- We support [implicit account](https://nomicon.io/DataStructures/Account.html#implicit-account-ids) creation which allows exchanges to create accounts without paying for transactions.
+- For exchanges, NEAR supports [implicit account](https://nomicon.io/DataStructures/Account.html#implicit-account-ids) creation which allows the creation of accounts without paying for transactions.
 - You can create an implicit account by following the steps in [this guide](/docs/roles/integrator/implicit-accounts).
-- Accounts must have enough tokens cover its storage. Storage cost per byte is 0.0001 NEAR and an account with one
-  access key must maintain a balance of at least 0.0182 NEAR. For more details, see [this section of the economics paper](https://near.org/papers/economics-in-sharded-blockchain/#transaction-and-storage-fees)
+- Accounts must have enough tokens cover its storage. Storage cost per byte is 0.0001 NEAR, and an account with one access key must maintain a balance of at least 0.0182 NEAR. For more details, see [this section of the economics paper](https://near.org/papers/economics-in-sharded-blockchain/#transaction-and-storage-fees).
```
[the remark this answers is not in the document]
mikedotexe: "typically costs"? I thought this was added by Bowen because it's precise. I think this is a constant, and don't want to add ambiguity

docs/roles/integrator/exchange-integration.md
mikedotexe: Gotta capitalize it properly to be https://httpie.io/ HTTPie 😇

docs/roles/integrator/exchange-integration.md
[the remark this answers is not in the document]
thisisjoshford: hmmm.. its there.. just not showing
Screen Shot 2021-05-27 at 9 28 19 PM

docs/roles/integrator/exchange-integration.md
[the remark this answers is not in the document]
thisisjoshford: Storage price actually can change but will never increase (as it would break a bunch of accounts) but can decrease if the price of NEAR goes up significantly and we want to do another 10x reduction in cost. I should change this wording from "typically" to "currently".

docs/roles/integrator/exchange-integration.md
thisisjoshford: done :)

docs/roles/integrator/exchange-integration.md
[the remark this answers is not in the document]
mikedotexe: Then this GitHub suggestion thing seems borked, good for an actual commit. I'm not gonna merge this, seems a glitch in this suggestion feature thingie

mikedotexe:
Hey @frol and @bowenwang1996 I think the RPC response has changed since publishing the documentation we're adding to here:
https://docs.near.org/docs/roles/integrator/exchange-integration

For example, in the link above we have a section I'm focusing on has a bolded title:

**Example of transfer from a lockup contract**

Here's a quick snippet from the top of the tx status query:

```bash
  "result": {
    "receipts": [
      {
        "predecessor_id": "evgeny.near",
        "receipt": {
```

Right off the bat this is different than what I'm seeing in my responses. `predecessor_id` seems to have been changed to `executor_id`. Is that correct? (Just making sure)
And I also don't see a `receipts` section.

But anyway, aside from that, here's my base understanding of how to approach this… It'll probably be useful for one of us to take a stab at a solution and you tell me if I'm wrong and what it should be instead.

**Goal**: we are trying to determine if a fungible token transfer was successful or not, and for how much. This will be ingested by exchanges as they use their own form of indexer that may or may not use the node indexer framework.

Here I have a transaction for an FT transfer. This is the simple kind. Let's not worry about `transfer_and_call` yet which is more complex.

Explorer:
https://explorer.testnet.near.org/transactions/CLScKgKkuUfjiR3xghy52Men7F95Q1uysJKfmQ1cnV4J

Get results via HTTPie:

`http post https://archival-rpc.testnet.near.org jsonrpc=2.0 id=ft-transfer method=tx params:='[ "CLScKgKkuUfjiR3xghy52Men7F95Q1uysJKfmQ1cnV4J", "ft.demo.testnet"]'`

Result (clickto expand/collapse):

<details>
  <summary>JSON response to ft_transfer</summary>
  
```json
{
  "id": "ft-transfer",
  "jsonrpc": "2.0",
  "result": {
    "receipts_outcome": [
      {
        "block_hash": "EMyNdkeYJQVJFPPqRoQxZh5i6XTs6bUh5CrTfATPh23K",
        "id": "DYgNBp8svBJLMJaRawJTW2DEVRcaYYVE5EJ1a6f47sdv",
        "outcome": {
          "executor_id": "ft.demo.testnet",
          "gas_burnt": 4770094374840,
          "logs": [
            "Transfer 1000000 from ft.demo.testnet to mike.testnet"
          ],
          "receipt_ids": [
            "AaKuP3PjbzBnu42WomNbSjsjrgvMXTQkyuvN5EY1uh6z"
          ],
          "status": {
            "SuccessValue": ""
          },
          "tokens_burnt": "477009437484000000000"
        },
        "proof": [
          {
            "direction": "Left",
            "hash": "2QzJXjSPDCYTw9z4CZqdN6mvzrKxxLb1sw6v7BrA2Ydy"
          },
          {
            "direction": "Right",
            "hash": "9MAZmB81AqSPJso6xomyNDziqF6imgUknzvkpJExXpLc"
          }
        ]
      },
      {
        "block_hash": "9eRSDDoJ8KjEVKfQtZjMEVoDNhubBKytoQYu84KYMXK1",
        "id": "AaKuP3PjbzBnu42WomNbSjsjrgvMXTQkyuvN5EY1uh6z",
        "outcome": {
          "executor_id": "ft.demo.testnet",
          "gas_burnt": 0,
          "logs": [],
          "receipt_ids": [],
          "status": {
            "SuccessValue": ""
          },
          "tokens_burnt": "0"
        },
        "proof": []
      }
    ],
    "status": {
      "SuccessValue": ""
    },
    "transaction": {
      "actions": [
        {
          "FunctionCall": {
            "args": "eyJyZWNlaXZlcl9pZCI6Im1pa2UudGVzdG5ldCIsImFtb3VudCI6IjEwMDAwMDAifQ==",
            "deposit": "1",
            "gas": 100000000000000,
            "method_name": "ft_transfer"
          }
        }
      ],
      "hash": "CLScKgKkuUfjiR3xghy52Men7F95Q1uysJKfmQ1cnV4J",
      "nonce": 49278550000006,
      "public_key": "ed25519:2JvNesazERwBkbFsgFsESvZ5bh78jTVEkjLP8tfLYbz2",
      "receiver_id": "ft.demo.testnet",
      "signature": "ed25519:2jQnRj5tvnPx4eoFTt54CgxsuVxXmxGn7DaZNqA8Yqpxshbdw52ZdkjiziiWoW51e4zkN81RYp7u3C25VwihGqUk",
      "signer_id": "ft.demo.testnet"
    },
    "transaction_outcome": {
      "block_hash": "EMyNdkeYJQVJFPPqRoQxZh5i6XTs6bUh5CrTfATPh23K",
      "id": "CLScKgKkuUfjiR3xghy52Men7F95Q1uysJKfmQ1cnV4J",
      "outcome": {
        "executor_id": "ft.demo.testnet",
        "gas_burnt": 2428055156040,
        "logs": [],
        "receipt_ids": [
          "DYgNBp8svBJLMJaRawJTW2DEVRcaYYVE5EJ1a6f47sdv"
        ],
        "status": {
          "SuccessReceiptId": "DYgNBp8svBJLMJaRawJTW2DEVRcaYYVE5EJ1a6f47sdv"
        },
        "tokens_burnt": "242805515604000000000"
      },
      "proof": [
        {
          "direction": "Right",
          "hash": "Em1LTLKfAp9CHsESqSu2GeQhahkNNFssR2jL8Ekeh6Vx"
        },
        {
          "direction": "Right",
          "hash": "9MAZmB81AqSPJso6xomyNDziqF6imgUknzvkpJExXpLc"
        }
      ]
    }
  }
}
```
</details>

Here's what I am guessing the exchanges should do to ensure the FT transfer was successful and for how much.

Look for `result` » `transaction_outcome` » `outcome` » (see if `SuccessReceiptId` is a key)
if `SuccessReceiptId` is not a key, this fungible token transfer has failed.
If it does have that key, get the value, which is a receipt ID.
Loop through `result` » `receipts_outcome` until you find an object that ID (from above) under the `id` key, then in that object check `outcome` » `status` » (see if `SuccessValue` is a key)
If `SuccessValue` is a key, fungible token transfer succeeded, if not, it failed.

To determine how *many* fungible tokens were transferred, look at:
`result` » `transaction` » `actions` » `FunctionCall` » `args`, then take the `args` and base64 decode it, that will give you a JSON payload and look for the `amount` key. It will contain a stringified number that represents the number of fungible tokens that were successfully transferred.

Good approach? Right? Wrong? Leave you comments below and don't forget to hit like and subscribe.

bowenwang1996:
> Right off the bat this is different than what I'm seeing in my responses. predecessor_id seems to have been changed to executor_id. Is that correct? (Just making sure)

No. `executor_id` only exists in execution outcomes and it is not part of `Receipt`.

volovyk-s:
@mikedotexe I have added `near-cli` and `JSON RPC` examples with responses to `Simple transfer example`.
+ I have used your explanation about how to check if the transaction was successfull. Please, review.
cc @thisisjoshford

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,993 @@ and we can find its outcome in `receipts_outcome`:
...
+      
+      {
+          "id": "storagebalancebounds",
+          "jsonrpc": "2.0",
+          "result": {
+              "block_hash": "Fy3mBqwj5nvUDha3X7G61kmUeituHASEX12oCASrChEE",
+              "block_height": 50060878,
+              "logs": [],
+              "result": [
+                  123,
```
mikedotexe, line 1154: I think just for readability, let's have all these bytes inline. so like

```
123, 34, 109…
```

instead of this really tall text

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,993 @@ and we can find its outcome in `receipts_outcome`:
...
+      Decoded result should look simmilar to:
+
+      ```bash
+      {
+        "min": "1250000000000000000000",
+        "max": "1250000000000000000000"
+      }
+      ```
+
+If there is not enough deposit for the storage or returned value is `null` - you should deposite more storage with the next command:
```
mikedotexe, line 1231: *deposit

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,993 @@ and we can find its outcome in `receipts_outcome`:
...
+            }
+          ]
+        },
+        "id": 125
+    }
+    ```
+
+You can get the same info later by the transaction hash from the previous call:
+
+  - using NEAR explorer: https://explorer.near.org
```
mikedotexe, line 1545: *Explorer

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +771,993 @@ and we can find its outcome in `receipts_outcome`:
...
+                  "SuccessValue": ""
+                }
+              }
+            }
+          ]
+        },
+        "id": 125
+    }
+    ```
+How to check if transfer was sucessfull?
```
mikedotexe, line 1739:
```suggestion
Was the fungible token transfer successful?
```

mikedotexe:
> > Right off the bat this is different than what I'm seeing in my responses. predecessor_id seems to have been changed to executor_id. Is that correct? (Just making sure)
> 
> No. `executor_id` only exists in execution outcomes and it is not part of `Receipt`.

Right, I think we're saying the same thing. I'm pointing out that this document is wrong now.
Screen Shot 2021-06-03 at 11 04 03 PM

I don't believe the above screenshot happens anymore, and unfortunately we have instructions to exchanges to look for `predecessor_id` here.

Screen Shot 2021-06-03 at 11 05 27 PM

I think we need to make another pass at the native token instructions in addition to the fungible token instructions.

docs/roles/integrator/exchange-integration.md
volovyk-s, line 1545:
```suggestion

  - using NEAR Explorer: https://explorer.near.org
```

frol:
> I don't believe the above screenshot happens anymore, and unfortunately we have instructions to exchanges to look for predecessor_id here.

@mikedotexe I believe that output is from `EXPERIMENTAL_tx_status` endpoint, we have not had a chance to stabilize it yet (it is absolutely safe to use this method, we will just replace the old `tx` method eventually)

docs/roles/integrator/exchange-integration.md
```diff
@@ -773,9 +841,1104 @@ and we can find its outcome in `receipts_outcome`:
...
+    https://explorer.mainnet.near.org/transactions/3MkWKbXVP8wyy4pBofELqiE1pwx7ie2v3SKCwaobNcEe
+    ''
+    ```
+
+  - with JSON RPC call:
+
+      ```bash
+      http post https://rpc.testnet.near.org jsonrpc=2.0 id=dontcare method=broadcast_tx_commit \
+          params:='["CwAAAHNlcmhpaS5uZWFyAAmQpgZcJM5nMc6f3tqmw/YI4eAvc84ZgsKMRRRzhY/6CQAAAAAAAAARAAAAYmVycnljbHViLmVrLm5lYXLLWPIiUOElkDF3u4hLAMJ0Sjeo1V338pDdHIp70va3ewEAAAACCwAAAGZ0X3RyYW5zZmVyKwAAAHsicmVjZWl2ZXJfaWQiOiJ2b2xvdnlrLm5lYXIiLCJhbW91bnQiOiIxIn0AQHoQ81oAAAEAAAAAAAAAAAAAAAAAAAAA7fDOZQt3zCtdS05Y8XaZFlwO/Gd5wkkNAHShzDiLQXk4Q4ixpraLPMJivs35PZD0gocXl1iGFbQ46NG3VllzCA=="]'
+      ```
```
frol, line 1270: @mikedotexe I fear this is not helpful as is. I had an idea in my backlog to write the docs about transaction construction, but so far it has not happened. I wonder how you came up with this transaction string... :thinking: 

I am using my `near-cli` (Rust implementation) to construct a transaction. Here is a completely offline version of the command (it does not require any network/internet connection and no other tool but `near-cli` binary itself):

```
$ near-cli construct-transaction \
    offline --block-hash 68GkSsoBuDMEHxELV3k3kqb7z6N8ij11Mj9XX4Wv97EV --nonce 123 \
    sender serhii.near \
    receiver berryclub.ek.near \
    add-action call-function --method-name ft_transfer --args '{"receiver_id": "volovyk.near", "amount": "1"}' --deposit '0 NEAR' --gas '100 TGas' skip \
    sign-private-key \
      --signer-secret-key 'ed25519:5Aj1vumfqLGNQjwsQD93fDzKZmhvBdAyCBkwrC83aKMqdyLHdcVyAFZ4W3Ej7Y4dsFPKNiVM99wfaSZReJSq5qz7' \
      --signer-public-key 'ed25519:8AJzPPSmGpNzGCUqsvz4X8zHomUEMqCM27ihgkKjQysX' \
    display
```

mikedotexe:
Task: fix the broken code and logic. I believe everything under the section **Example of transfer from a lockup contract** has been outdated for a while. We also have logic explaining to exchanges:

>they only need to filter receipts that only contain Transfer action and whose predecessor_id is not system

There may be more that's wrong, but for sure it's not `predecessor_id` when we use NEAR CLI's `tx-status`

I think that's because NEAR CLI uses the `tx` instead of `EXPERIMENTAL_tx_status`. So either we need to meet with @frol about expectations for this getting a different name, or being backwards compatible, and if we should change NEAR CLI to use it. OR we need to remove all NEAR CLI references here. We just have to make sure exchanges can follow these instructions and can see the same thing we're showing them they should see, which is currently not the case.

docs/roles/integrator/exchange-integration.md
mikedotexe: Gotcha, so the way we were dealing with this is to have links to a page in docs where you create transactions.

>To follow this guide, please check the step by step instructions on how to create a transaction first.

Which links to:
https://docs.near.org/docs/tutorials/create-transactions#low-level----create-a-transaction

What do you think?

frol: Yeah, providing the link definitely makes sense. Maybe we want to mentions some details about the transaction that needs to be sent in this particular case, which is:

```
Transaction {
        block_hash: `68GkSsoBuDMEHxELV3k3kqb7z6N8ij11Mj9XX4Wv97EV`,
        signer_id: "serhii.near",
        public_key: ed25519:3oN25HE9tf9YUGXQyXCFB2rEtpomijG8ghVV2JvrA4zJ,
        nonce: 123,
        receiver_id: "berryclub.ek.near",
        actions: [
            FunctionCall(
                FunctionCallAction {
                    method_name: ft_transfer,
                    args: `{"receiver_id": "volovyk.near", "amount": "1"}`,
                    gas: 100000000000000,
                    deposit: 0,
                },
            ),
        ],
    }
```

You may format it as JSON and only highlight `signer_id`, `receiver_id`, and `actions`.

mikedotexe: Thanks, Frol, I've pushed up changes with this idea in mind. I think it's a great idea.

mikedotexe:
hey @thisisjoshford I see you have a change requested, but think it might be a comment. All good?
Everyone else: ready to review

frol (approved):

volovyk-s (approved):

thisisjoshford (approved):

PR #724, opened 2021-05-27
Instructions for Debug testnet contract locally

ailisp:
This is mainly the first part, obtain enough information to setup the sandbox node. The second part, write a test to replay the transaction, is same as run a test on a sandbox node as showing in #706 . I'm going to write a full example on a reasonably complex contract, with cross contract calls on testnet, and steps cover first&second part on how to replay the transaction on sandbox node.

Update 6/15:
I just add a section to use script to automatic obtain enough information for the sandbox node and give an example of step-by-step reproduce testnet contract locally


docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+sidebar_label: Debug Locally
+---
+
+After contracts are deployed on testnet or mainnet, you might still experience corner case of:
+
+- contract execution logic does not happen in expected way;
+- storage usage is higher than expected;
+- gas cost is higher than expected.
+
+In these cases, you may want to capture the contract state and function call arguments to debug in the local sandbox. You'll make incremental changes, test locally until everything works as expected. You will also likely want to keep the test fixtures to avoid regression in future.
```
thisisjoshford, line 13:
```suggestion
In these cases you may want to capture the contract state and function call arguments to debug in the local sandbox. You'll make incremental changes and test locally until everything works as expected. You will also likely want to keep the test fixtures to avoid regression in future.
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Obtain States and Other Parameters from Testnet/Mainnet
+
+The following things determined the contract call results:
```
thisisjoshford, line 17:
```suggestion
The following things determine the contract call results:
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+- Contract code and state;
+- Node's genesis config;
+- Context data when the transaction happens, like current epoch, block height, gas price etc;
+- Caller (signer) of the transaction and arguments.
```
thisisjoshford, line 22:
```suggestion
- Contract code and state.
- Node's genesis config.
- Context data when the transaction happens like current epoch, block height, gas price, etc.
- Caller (signer) of the transaction and arguments.
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+We'll need to obtain all this information to reproduce the contract call locally. This information can be divide into three categories, each of them is configured in a different way:
```
thisisjoshford, line 24:
```suggestion
We'll need to obtain all this information to reproduce the contract call locally. This information can be divided into three categories, each of which is configured differently:
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+- Genesis config
+- State records
+- Parameter provided when sending transaction
```
thisisjoshford, line 28:
```suggestion
1. Genesis config
2. State records
3. Parameter provided when sending transaction
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Genesis config
+
+The context data when transaction happens, like current epoch, block height, gas price etc. and testnet/mainnet specific configuration such as block time, gas amount for different kind of actions all come into this category. This parameters can only be set by start a local sandbox node from a modified version of genesis file.
```
thisisjoshford, line 32:
```suggestion
The context data when a transaction happens such as the current epoch, block height, gas price, etc., as well as `testnet`/`mainnet` specific configurations (block time / the gas amount for different kinds of actions) are all part of this category. This parameter can only be set by starting a local sandbox node from a modified version of the genesis file.
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+First thing you need is fetch testnet/mainnet genesis config. This can be obtained by [Genesis Config RPC](https://docs.near.org/docs/api/rpc#genesis-config). Use response file to replace sandbox's genesis.json. Now we have the same configuration as the testnet/mainnet in sandbox. But there's still something different: protocol version, block height and block timestamp. This is because genesis indicates the start of the most recent hard fork of testnet/mainnet, but many blocks and time has passed since the genesis. There's also protocol version upgrades since genesis. If your contract logic doesn't depend on these parameters replay contract method on such parameters, sandbox node should give you accurate logic, gas and storage simulation. However if you do require info like certain epoch and block height, you can modify `"genesis_height"`, `"protocol_version"` on genesis.json.
```
thisisjoshford, line 33:
```suggestion
The first thing you'll need to do is fetch the `testnet`/`mainnet` genesis config. This can be obtained by the [Genesis Config RPC](https://docs.near.org/docs/api/rpc#genesis-config). Use the response file to replace the sandbox's `genesis.json`. Once complete, you will have the same configuration as `testnet`/`mainnet` in the sandbox aside from the protocol version, block height, and block timestamp. This is because genesis indicates the start of the most recent hard fork of `testnet`/`mainnet` but many blocks have been produced, protocol versions released, and time has passed the blockchain genesis. If your contract logic doesn't depend on these parameters the sandbox node should give you accurate logic, gas, and storage simulation. However, if you do require info like certain epoch and block height details you can modify `"genesis_height"` and `"protocol_version"` on your `genesis.json` file.
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+Change `"genesis_time"` won't have effect. There's going to be a time travel RPC for you to tweak block timestamps. Also note a transaction cannot be controlled to be precisely scheduled of a given height and you should not have absolute block number in test logics. This is because it's not controllable in a real node so your contract logic shouldn't rely on that. More common you want to depend on how many blocks or times have passed, that's a relative number. And you may want to base on the parameters at the time block produced, such as validator set, protocol feature availability etc., these are tied to epoch rather than block, so you don't have to strictly depend on a given block number.
```
thisisjoshford, line 35:
```suggestion
Simply changing `"genesis_time"` won't have an effect. In addition, there's going to be a time travel RPC for you to tweak block timestamps. Also, note a transaction cannot be controlled to be precisely scheduled of a given height and you should not have absolute block number in test logics. This is because it's not controllable in a real node so your contract logic shouldn't rely on that. More commonly, you most likely want to depend on how many blocks or how much time has passed. This is a relative number and you may want to base on the parameters at the time the block was produced, such as a validator set, protocol feature availability, etc. These are tied to [epoch](/docs/concepts/epoch) rather than the individual block so you don't have to strictly depend on a given block number.
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+### State Records
+
+Every state stored on NEAR Blockchain including contract code, contract states, accounts, access keys are in the form of state records. To run a node having equivalent state as a testnet/mainnet node you need to apply same state records as the testnet/mainnet. You don't need to apply all states which is many gigabytes, but only those contract code/state and account keys that were used in the transaction. Once you have these state records, you can apply them by either place them in the `"records"` section of genesis before starting sandbox, or use `sandbox_patch_state` RPC to add them dynamically after starting sandbox. Let's see how state records can be obtained:
```
thisisjoshford, line 39:
```suggestion
Every state stored on the NEAR Blockchain including contract code, contract states, accounts, and access keys is in the form of state records. To run a node having an equivalent state as a `testnet`/`mainnet` node you need to apply the same state records as `testnet`/`mainnet`. You don't need to apply all states which can be many gigabytes but only those contract code/state and account keys that were used in the transaction. Once you have these state records you can apply them by either placing them in the `"records"` section of genesis _before_ starting the sandbox or use `sandbox_patch_state` RPC to add them dynamically _after_ starting the sandbox. Let's see how these state records can be obtained:
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+For contract code, Most likely you already have the contract code you've deployed. But in case you don't, you can fetch it by `query` RPC:
```
thisisjoshford, line 41:
```suggestion
**Contract code:** 

Most likely you already have the contract code you've deployed but in case you don't you can fetch it by `query` RPC:
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+```
+curl https://rpc.testnet.near.org -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0", "id":1, "method":"query", "params":{"request_type":"view_code","block_id": block-height-or-hash,"account_id":"contract-account"}}'
+```
+
+You can also replace the url to `https://rpc.mainnet.near.org` to fetch code from mainnet. This is same for all following RPC requests in this article.
```
thisisjoshford, line 47:
```suggestion
You can also replace the url to `https://rpc.mainnet.near.org` to fetch code from `mainnet`. This is same for all following RPC requests in this article.
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+For contract state, there's a few ways of getting it. If your contract state is less than 50Kib, you can fetch it with `query` RPC:
```
thisisjoshford, line 49:
```suggestion
**Contract state:** 

There are a few ways of getting contract state. If your contract state is less than 50Kib you can fetch it with `query` RPC:
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+```
+curl https://rpc.testnet.near.org -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0", "id":1, "method":"query", "params":{"request_type":"view_state","block_id":block-height-or-hash,"account_id":"contract-account","prefix_base64":""}}'
+```
+
+Otherwise, you need to run a testnet or mainnet node with `nearup`, and wait it fully synced. Then you can either use `state-viewer` to aggreggate entire state to a new genesis, or use `state-viewer` to export state from only specified contracts.
```
thisisjoshford, line 55:
```suggestion
Otherwise, you need to run a `testnet` or `mainnet` node with `nearup` and wait for it fully synced. Then you can either use `state-viewer` to aggregate the entire state to a new genesis or use `state-viewer` to export the state from only specified contracts.
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+For account and access key, you can use [view account](https://docs.near.org/docs/api/rpc#view-account) and [view access keys](https://docs.near.org/docs/api/rpc#view-access-key-list) RPCs, respectively.
```
thisisjoshford, line 57:
```suggestion
**Account and Access Key:**

For account and access key, you can use [view account](https://docs.near.org/docs/api/rpc#view-account) and [view access keys](https://docs.near.org/docs/api/rpc#view-access-key-list) RPCs, respectively.
```

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Parameter provided when sending transaction
+
+This can be found by inspect the transaction information on Explorer. This is majorly the function call arguments, gas attached, deposit attached and the signer of the transaction. If you do not control keys of the transaction caller (signer), you can use the patch_state RPC to add your public key to account.
```
thisisjoshford, line 61:
```suggestion
This can be found by inspecting the transaction information on [NEAR Explorer](/docs/tools/near-explorer). This is majorly the function call arguments, gas attached, deposit attached, and the signer of the transaction. If you do not control the keys of the transaction caller (signer) you can use the `patch_state` RPC to add your public key to the account.
```

thisisjoshford (changes requested):
Fixed some grammar and added some wording suggestions :)

thisisjoshford (approved):

ailisp:
>  I just add a section to use script to automatic obtain enough information for the sandbox node and give an example of step-by-step reproduce testnet contract locally

@thisisjoshford I added ^ two sections, ptal :) Thanks

docs/develop/contracts/debug-locally.md
```diff
@@ -0,0 +1,178 @@
...
+
+### State Records
+
+Every state stored on the NEAR Blockchain including contract code, contract states, accounts, and access keys is in the form of state records. To run a node having an equivalent state as a `testnet`/`mainnet` node you need to apply the same state records as `testnet`/`mainnet`. You don't need to apply all states which can be many gigabytes but only those contract code/state and account keys that were used in the transaction. Once you have these state records you can apply them by either placing them in the `"records"` section of genesis _before_ starting the sandbox or use `sandbox_patch_state` RPC to add them dynamically _after_ starting the sandbox. Let's see how these state records can be obtained:
```
thisisjoshford, line 150:
```suggestion
Every state stored on the NEAR Blockchain including contract code, contract states, accounts, and access keys is in the form of state records. To run a node having an equivalent state as a `testnet`/`mainnet` node you need to apply the same state records as `testnet`/`mainnet`. You don't need to apply all states, which can be many gigabytes, but only those contract code/state and account keys that were used in the transaction. Once you have these state records you can apply them by either placing them in the `"records"` section of genesis _before_ starting the sandbox or use `sandbox_patch_state` RPC to add them dynamically _after_ starting the sandbox. Let's see how these state records can be obtained:
```

thisisjoshford (approved):
Left some suggestions. :) Other than that, looks good to go!

ailisp:
Did a final proof read on render, looks suggestions are applied and overall looks good

PR #746, opened 2021-06-09
Postmortem about hash collisions

telezhnaya:
We need to find the proper place for it, I couldn't decide whether it should be in the docs at all.

### DIFF
--- a/docs/community/hash_collision_postmortem.md
+++ b/docs/community/hash_collision_postmortem.md
@@ -1,0 +1,35 @@
+## POSTMORTEM June 9th, 2021
+
+### Quick summary
+
+On December 2, 2020, we noticed the determined way of creating [transaction hash collisions on Testnet](https://github.com/near/nearcore/security/advisories/GHSA-2v6r-g342-282f), then also on Mainnet.
+
+It still affects the data representation in some services such as [Near Explorer](https://explorer.near.org).
+
+[The fix](https://github.com/near/nearcore/pull/4064) was merged on 11 of March 2021 and released in [1.19.0](https://github.com/near/nearcore/releases/tag/1.19.0), so new examples of reproducible hash collisions will not appear again.
+
+### In details
+
+There was a possibility to apply the same transaction two or more times.
+Such transactions are logically different objects, but they have the same hash.
+It means that now, in the general case, the transaction hash does not fully identify the transaction. 
+
+We observe some limitations because of that:
+- It's not possible to get the non-first appearance of transaction via [RPC](https://docs.near.org/docs/api/rpc#transaction-status).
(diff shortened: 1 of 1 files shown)


docs/community/hash_collision_postmortem.md
```diff
@@ -0,0 +1,35 @@
...
+On December 2, 2020, we noticed the determined way of creating [transaction hash collisions on Testnet](https://github.com/near/nearcore/security/advisories/GHSA-2v6r-g342-282f), then also on Mainnet.
+
+It still affects the data representation in some services such as [Near Explorer](https://explorer.near.org).
+
+[The fix](https://github.com/near/nearcore/pull/4064) was merged on 11 of March 2021 and released in [1.19.0](https://github.com/near/nearcore/releases/tag/1.19.0), so new examples of reproducible hash collisions will not appear again.
+
+### In details
+
+There was a possibility to apply the same transaction two or more times.
+Such transactions are logically different objects, but they have the same hash.
```
bowenwang1996, line 14: Not sure what you mean by logically different objects -- they are the *same* transaction

telezhnaya: In a normal situation, the transaction could be applied only once. We have the situation when it is applied twice or more times. I needed to speak about different appearances. Sorry if that confuses you. 
Is it clear now what I wanted to say?
Could you please correct my wording?

bowenwang1996: I think stating that a transaction should only be applied once is probably better

telezhnaya: This wording does not explain our case. It should be applied only once, but it's not true in the real world.

bowenwang1996: Well what I meant is that it should be applied only once, but due to the issue we had, some transaction was applied more than once.

frol: @bowenwang1996 It seems we lost the shared context. This page is a postmortem about the transactions being included in two different blocks/chunks, and thus applied twice, and thus having two execution outcomes, got converted to two different receipts, and the following execution happened as if those transactions are different, but on Indexer for Explorer side, we cannot really disambiguate this scenario unless we introduce unnecessary redundancy which will confuse people even more than that.

thisisjoshford:
@telezhnaya is this PR still valid? We are looking to migrate to docusarus v2 and clean up all stale PRs.

telezhnaya:
@thisisjoshford yes, unfortunately, we still do not have any solution here

frol:
@telezhnaya Let's move this from the docs to `near-indexer-for-explorer` issue/discussions since I don't think we have any place on the docs for this type of communication at the moment.

PR #751, opened 2021-06-10
Fix creating key store from file system in near-api-js quick reference

think-in-universe:
### Description

As reported by developers, the [`keyStore` section](https://docs.near.org/docs/api/naj-quick-reference#key-store) in near-api-js quick reference contains mistakes of how to create `keyStore` from file system. 

The examples in cookbook seem to be correct: https://docs.near.org/docs/api/naj-cookbook#create-account

I fixed this with two examples of loading `keyStore` from a file and from the credentials directory.

### DIFF
--- a/docs/api/naj-quick-reference.md
+++ b/docs/api/naj-quick-reference.md
@@ -71,7 +71,8 @@ const homedir = require("os").homedir();
 
 const ACCOUNT_ID = "near-example.testnet";  // NEAR account tied to the keyPair
 const NETWORK_ID = "testnet";
-const KEY_PATH = `/.near-credentials/${NETWORK_ID}/${ACCOUNT_ID}.json`;
+// path to your custom keyPair location (ex. function access key for example account)
+const KEY_PATH = '/.near-credentials/near-example-testnet/get_token_price.json';
 
 const credentials = JSON.parse(fs.readFileSync(homedir + KEY_PATH));
 const keyStore = new keyStores.InMemoryKeyStore();
--- a/docs/api/naj-quick-reference.md
+++ b/docs/api/naj-quick-reference.md
@@ -48,8 +48,9 @@ const keyStore = new keyStores.BrowserLocalStorageKeyStore();
 <!--Using Credentials Directory-->
 
 ```js
-// creates keyStore from credentials directory
-// you should have login the account with NEAR CLI: https://docs.near.org/docs/tools/near-cli#near-login
+// creates a keyStore that searches for keys in .near-credentials
+// requires credentials stored locally by using a NEAR-CLI command: `near login` 
+// https://docs.near.org/docs/tools/near-cli#near-login
 
 const { keyStores } = nearAPI;
 const homedir = require("os").homedir();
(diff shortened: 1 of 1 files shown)


docs/api/naj-quick-reference.md
```diff
@@ -45,15 +45,30 @@ const { keyStores } = nearAPI;
...
 // you will need to pass the location of the .json key pair
 
-const { keyStores } = nearAPI;
-const KEY_PATH = "~./near-credentials/testnet/example-account.json";
-const keyStore = new keyStores.UnencryptedFileSystemKeyStore(KEY_PATH);
+const { keyStores, KeyPair } = nearAPI;
+const KEY_PATH = "~/.near-credentials/testnet/example-account.json";
+const credentials = JSON.parse(fs.readFileSync(KEY_PATH));
+const keyStore = new keyStores.InMemoryKeyStore();
+keyStore.setKey(networkId, appName, KeyPair.fromString(credentials.private_key));
```
thisisjoshford, line 71:
```suggestion
const { KeyPair, keyStores } = require("near-api-js");
const fs = require("fs");
const homedir = require("os").homedir();

const ACCOUNT_ID = "near-example.testnet";  // NEAR account tied to the keyPair
const NETWORK_ID = "testnet";
const KEY_PATH = "/.near-credentials/testnet/near-example.testnet.json";

const credentials = JSON.parse(fs.readFileSync(homedir + KEY_PATH));
const keyStore = new keyStores.InMemoryKeyStore();
keyStore.setKey(NETWORK_ID, ACCOUNT_ID, KeyPair.fromString(credentials.private_key));
```

Nice catch! Wasn't able to get my machine to read the json file by passing the KEY_PATH using '~' even though I am on a mac... so decided to include `homedir` in this example as well.

docs/api/naj-quick-reference.md
```diff
@@ -45,15 +45,30 @@ const { keyStores } = nearAPI;
 const keyStore = new keyStores.BrowserLocalStorageKeyStore();
 ```
 
+<!--Using Credentials Directory-->
+
+```js
+// creates keyStore from credentials directory
+// you should have login the account with NEAR CLI: https://docs.near.org/docs/tools/near-cli#near-login
```
thisisjoshford, line 52:
```suggestion
// creates a keyStore that searches for keys in .near-credentials
// requires credentials stored locally by using a NEAR-CLI command: `near login` 
// https://docs.near.org/docs/tools/near-cli#near-login
```

thisisjoshford (approved):
thanks for the fix! left some comments / suggestions :)

docs/api/naj-quick-reference.md
think-in-universe: ah, yes. that's right. thanks!

docs/api/naj-quick-reference.md
```diff
@@ -71,7 +71,7 @@ const homedir = require("os").homedir();
 
 const ACCOUNT_ID = "near-example.testnet";  // NEAR account tied to the keyPair
 const NETWORK_ID = "testnet";
-const KEY_PATH = "/.near-credentials/testnet/near-example.testnet.json";
+const KEY_PATH = `/.near-credentials/${NETWORK_ID}/${ACCOUNT_ID}.json`;
```
thisisjoshford, line 74: I'm wondering if it would make sense to have the `KEY_PATH` be something unique that would indicate the special use case of pointing directly to a file. If a user was using the default keyPair for an account in the default directory of that network... they should just use the "Using Credentials Directory" example.

Maybe we can change this to highlight that you would only use this if you wanted to use a special key pair to sign this transaction... maybe a special function access key or something of the like... What do you think @think-in-universe ?

Something like:



```suggestion
// path to your custom keyPair location (ex. function access key for example account)
const KEY_PATH = '/.near-credentials/near-example-testnet/get_token_price.json';
```

think-in-universe: good idea. we do have developers who want to test with function call key in this way.

PR #758, opened 2021-06-15
Update compile-and-run-a-node.md

TrevorJTClarke:
added note about disk needs


khorolets (approved):
Great one! Thanks for that!

PR #773, opened 2021-07-05
Fix wget call

Longarithm:
`wget DST SRC` didn't work for me on both Mac and Ubuntu, but command `wget SRC -P DST_DIR` worked.

Proof: https://stackoverflow.com/questions/1078524/how-to-specify-the-download-location-with-wget


janewang (approved):

chefsale (approved):
Thanks for fixing this.

PR #782, opened 2021-07-14
Update documentation to mention `make neard`

mina86:
We are planning to disable LTO in Cargo.toml which means that
`cargo build --release` will produce less optimised binaries (but
do it faster).  Point at `make neard` in the documentation to make
sure that anyone intending to run node in production gets best
optimised version as before.

The `make neard` target produces the `neard` binary but it has been
added only recently† and is not available when building older
releases.  Because of that, when documentation mentions checking out
non-HEAD commits points at `make release` instead.  It builds a few
more targets so it is slower but it does produce `neard`.

Note that docs/community/contribute/nearcore.md file has been left
unchanged because it describes development process (rather than
running the node in production) and in that case we want people to
build via `cargo build` et al which is much faster‡.

† https://github.com/near/nearcore/pull/4510
‡ https://github.com/near/nearcore/pull/4507

Issue: https://github.com/near/nearcore/issues/4403

### DIFF
--- a/docs/develop/node/archival/run-archival-node-without-nearup.md
+++ b/docs/develop/node/archival/run-archival-node-without-nearup.md
@@ -50,7 +50,6 @@ First, clone the [`nearcore` repository](https://github.com/near/nearcore).
 ```bash
 $ git clone https://github.com/near/nearcore
 $ git fetch origin --tags
-$ cd nearcore
 ```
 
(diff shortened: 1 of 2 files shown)


janewang:
Thanks for updating the docs! 👍  Our documentation isn't DRY and we have another page to update 🙏 

develop/node/archival/run-archival-node-without-nearup.md
L67:$ cargo build --release --package neard --bin neard
L186:$ cargo build --release --package neard --bin neard

mina86:
Done.

janewang (approved):

PR #793, opened 2021-08-05
Add link to near-examples repo for AS cross-contract calls

mikedotexe:
This is a resource folks have found helpful and I believe we'll want to link to it somewhere.


thisisjoshford (approved):
Awesome, thanks @mikedotexe !

PR #800, opened 2021-08-08
Correct link in block quote

Arthur-Milchior:
Both on https://docs.near.org/docs/tokens/token-custody and in github, the content of blockquote is NOT markdown but html. So requires HTML formatting


bucanero (approved):
You're right, thanks for fixing this broken link 👍

thisisjoshford (approved):
Nice catch! Thanks!

PR #803, opened 2021-08-11
Restore data backup to expected directory

jbryce:
When executing the bash commands as documented, all of the files in the backup archive end up extracted into the top level of the near home directory. On first start of neard, it then creates a new data structure in near home/data and begins to sync with peers from the beginning. The updates in this commit extract the archive into the expected /data directory and allow the synchronization to start from the last back up.

### DIFF
--- a/docs/develop/node/validator/compile-and-run-a-node.md
+++ b/docs/develop/node/validator/compile-and-run-a-node.md
@@ -199,9 +199,10 @@ or
 2. Run the following commands:
 
 ```bash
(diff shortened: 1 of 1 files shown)


thisisjoshford:
@chefsale @mhalambek is this PR still valid? We are looking to migrate to docusaurus v2 and want to clean up all stale PRs.

bucanero:
Node-related documentation has been moved to [node-docs](https://github.com/near/node-docs)

PR #805, opened 2021-08-11
truncate sandbox docs, link to near-runner

chadoh:
I was going to update the page about NEAR Sandbox with up-to-date info about the newer, better way to write tests with [near-runner](https://github.com/near/runner)

But then it seemed like I was just going to copy-paste its readme into this page.

I generally think:

* We should avoid copying & pasting most things, definitely readmes
* The further documentation lives from the code it documents, the higher-level it should be

And the readme for near-runner is already quite high-level and approachable!

So I'm not sure what to put on this page anymore. @thisisjoshford what do you think?


bucanero:
I agree that providing information in the repository's README is good when working with open-source dev projects. (My rule of thumb is to check that all the information can fit in a one-page README)

Regarding the sandbox, I was checking https://github.com/near/runner/blob/main/README.md, and the only information I see is a Youtube video link. I'm not sure if that's the right README?

In any case, I think videos are nice, but they can't replace all the documentation. Also, videos are harder to link and can't be scanned by search engines, so we might miss users that are trying to find stuff about the sandbox.
For example, in the current docs (https://docs.near.org/docs/develop/contracts/sandbox) you have steps the user can follow, code examples, and bash commands; those are nice to have since the user can easily copy&paste. 

From what I understand, when we update to the new Docusaurus we'll be able to inject/integrate content directly from GitHub into our docs, so we can provide the information on the repo's README, and also have it on docs.near.org.

chadoh:
Sorry, `near/runner` is now an index page linking to language-specific libraries. The more relevant README is [the one for near/runner-js](https://github.com/near/runner-js#readme)

bucanero:
Hello @chadoh 

I was reviewing the docs from `runner-js`, `sandbox`, and this PR. The Readme on `runner-js` looks nice! 👍 
I submitted a small PR on the `sandbox` repository with some minor suggestions.

I only have one question left, regarding `sandbox_patch_state`. On the current page (docs.near.org) we have some notes about "Patch state on the fly", but I don't see any mention of it in the new Readmes.
Is this sandbox-only feature still available? (patch state). If it's still supported, we should mention it somewhere. If it's no longer available, then you can skip my comment. 😄 

Let me know, thank you!

chadoh:
Great catch, @bucanero! This info is still relevant and was only documented in near-runner-js via its tests, which is not nearly as helpful as the info I'm here proposing we remove from docs.near.org. I've now added this info [to near-runner-js](https://github.com/near/runner-js/pull/41) and [to near-sandbox](https://github.com/near/sandbox/pull/11).

bucanero (approved):
LGTM 👍

bucanero:
hello @chadoh, is this PR still a draft? or should I review and merge? let me know, thanks!

PR #813, opened 2021-08-17
Add links to NEAR's Wiki

bucanero:
This PR adds links to the new Wiki (https://github.com/near/devrel/issues/255):

- Redirects `Staking` header link to the wiki
- Adds footer link to Wiki


thisisjoshford (approved):
Looks good! I think we should make the Wiki more visible on the footer though... I'll push a quick edit.

bucanero:
> Looks good! I think we should make the Wiki more visible on the footer though... Can you move it to line 58 & 1/2? So its right under "Examples"?

Done 👍 `Wiki` link moved under `Examples`

PR #817, opened 2021-08-19
Remove AssemblyScript for financial warning

vgrichina:


nearmax (changes requested):
In the past we have identified the criteria for removing this label:
* AssemblyScript compiler needs to have 1.0 stable version;
* AS needs to be used in production by several major organizations, e.g. someone from FAANG;
* AS code needs to have full test coverage.

None of it has been satisfied yet and therefore we cannot remove this notice.

CC @DiscRiskandBisque @jberrytech

PR #822, opened 2021-08-23
[DOC] Clarify Wasm, VMContext, Env. Add Rust Transaction example

RNHTTR:
**Is your feature request related to a problem? Please describe.**
It's difficult to understand the relationship between wasm, VMContext, and env, and it's really hard to find how to send NEAR from user to user. 

**Describe the solution you'd like**
A specific page for Wasm, VMContext, and env and a code example for using `Promise` to transfer funds between two users. I could use help describing wasm and vmcontext.

### DIFF
--- a/docs/develop/contracts/rust/near-sdk-rs.md
+++ b/docs/develop/contracts/rust/near-sdk-rs.md
@@ -138,16 +138,21 @@ impl Default for StatusMessage {
 
 * **Payable methods.** We can allow methods to accept token transfer together with the function call. This is done so that contracts can define a fee in tokens that needs to be payed when they are used. By the default the methods are not payable and they will panic if someone will attempt to transfer tokens to them during the invocation. This is done for safety reason, in case someone accidentally transfers tokens during the function call. 
 
-To declare a payable method simply use `#[payable]` decorator:
+To declare a payable method simply use `#[payable]` decorator. The following example transfers funds from the [signer of the contract](https://docs.rs/near-sdk/latest/near_sdk/env/fn.signer_account_id.html) to `receiver.testnet` using [`Promise`](https://docs.rs/near-sdk/latest/near_sdk/struct.Promise.html):
 ```rust
-use near_sdk::payable;
+use near_sdk::{
+    AccountId,
+    payable,
+    Promise,
+}
 
 #[payable]
-pub fn my_method(&mut self) {
-...
+pub fn send_funds(&mut self) {
+    let receiver_id: AccountId = "receiver.testnet".to_string();
+    let amount: u128 = 1_000_000_000_000_000_000_000_000;
+    Promise::new(receiver_id).transfer(amount);
 }
 ```
-
 
 ## Pre-requisites
 To develop Rust contracts you would need to:
--- a/docs/develop/contracts/rust/wasm-vmcontext-and-env.md
+++ b/docs/develop/contracts/rust/wasm-vmcontext-and-env.md
@@ -1,0 +1,22 @@
+---
+id: wasm-vmcontext-and-env
+title: Understanding the Importance of WebAssembly, VMContext, and env
+sidebar_label: Wasm, VMContext, and env
+---
+
+# Understanding the Importance of WebAssembly, VMContext, and env
+## WebAssembly
+NEAR compiles smart contracts to [WebAssembly](https://webassembly.org/), or Wasm. Wasm helps NEAR protocol transactions to be fast, safe, and open.
+
+TODO
+
+## VMContext
+TODO
+
+## env
+[`near-sdk::env`](https://docs.rs/near-sdk/latest/near_sdk/env/index.html) makes blockchain specific methods available to the smart contract. These methods include (but aren't limited to):
+* [`account_balance`](https://docs.rs/near-sdk/latest/near_sdk/env/fn.account_balance.html)
+* [`current_account_id`](https://docs.rs/near-sdk/latest/near_sdk/env/fn.current_account_id.html)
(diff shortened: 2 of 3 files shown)


RNHTTR:
Also should include documentation providing context around contract initialization, re-deploying a contract and what happens to state, and how to access blockchain storage

RNHTTR:
> Also should include documentation providing context around contract initialization, re-deploying a contract and what happens to state, and how to access blockchain storage

Also some additional commentary on the idea that ["All data stored on the NEAR blockchain is done in key / value pairs."](https://docs.near.org/docs/concepts/data-storage#overview). It'd be useful to describe in `docs/concepts/data-storage` how to initialize a `Vector`, `UnorderedMap`, etc., and what it means to pass `prefix: S` / `IntoStorageKey`

docs/develop/contracts/rust/wasm-vmcontext-and-env.md
```diff
@@ -0,0 +1,22 @@
...
+TODO
+
+## VMContext
+TODO
+
+## env
+[`near-sdk::env`](https://docs.rs/near-sdk/latest/near_sdk/env/index.html) makes blockchain specific methods available to the smart contract. These methods include (but aren't limited to):
+* [`account_balance`](https://docs.rs/near-sdk/latest/near_sdk/env/fn.account_balance.html)
+* [`current_account_id`](https://docs.rs/near-sdk/latest/near_sdk/env/fn.current_account_id.html)
+* [`is_valid_account_id`](https://docs.rs/near-sdk/latest/near_sdk/env/fn.is_valid_account_id.html)
```
austinabell, line 20: why do we want to advertise this? If anything we should note that AccountId's are being validated through its type (as of v4) and not require using this API which is very unergonomic and should only be used in edge cases

docs/develop/contracts/rust/near-sdk-rs.md
```diff
@@ -138,17 +138,22 @@ impl Default for StatusMessage {
...
+    Promise,
+}
 
 #[payable]
-pub fn my_method(&mut self) {
-...
+pub fn send_funds(&mut self) {
+    let receiver_id: AccountId = "receiver.testnet".to_string();
+    let amount: u128 = 1_000_000_000_000_000_000_000_000;
+    Promise::new(receiver_id).transfer(amount);
```
austinabell, line 153: I'm confused, this doesn't make sure that the necessary funds were sent to the contract, isn't this a bad pattern to use as an example?

austinabell:
Thanks for the contribution @RNHTTR, a lot of the rust documentation is being updated/added here https://www.near-sdk.io/ which includes roughly the changes you're proposing here. https://github.com/near/sdk-docs is the repo, for reference.

RNHTTR:
> Thanks for the contribution @RNHTTR, a lot of the rust documentation is being updated/added here https://www.near-sdk.io/ which includes roughly the changes you're proposing here. https://github.com/near/sdk-docs is the repo, for reference.

Aha! That looks great, and like you said, covers a lot of what I'm trying to capture. Ultimately, as I learn about NEAR & the rust sdk, I wanted to capture some of the things that I found challenging/confusing and be proactive in updating the documentation accordingly.

Not sure this PR is the best forum to continue discussion, but I've found the following difficult to follow:

- What is `env`, what are its properties/methods, and why are these significant?
  - Where does `env` even come from? Can these be manipulated from the front-end when the contract is called?
  - Who exactly is `env::current_account_id`, `env::predecessor_account_id`, `env::signer_account_id`, and how are these set?
- How can a contract let user A send NEAR to user B? This is what I was trying to capture with the `send_funds` method, but it looks what I was trying to convey was nicely captured [in the docs you shared](https://www.near-sdk.io/promises/token-tx).
- Understanding the on-chain storage components of a contract. I thought that when a contract was deployed, each time the contract was called, the `new` method would be called, and the properties in `new` would be set each time the contract was called.
  - On a similar note, seeing code like `LookupMap::new(b"my_property".to_vec())` was really confusing to me. It'd be nice if there was some additional clarification on the key / value paradigm for on-chain storage, outlined in the [overview here](https://docs.near.org/docs/concepts/data-storage#overview). Now I understand that `"my_property".to_vec()` sets the key for the storage object on-chain. At least I think I understand it :)

Let me know if this makes sense or not. Thanks!

austinabell:
> Not sure this PR is the best forum to continue discussion, but I've found the following difficult to follow:
To

> 
> * What is `env`, what are its properties/methods, and why are these significant?
>   
>   * Where does `env` even come from? Can these be manipulated from the front-end when the contract is called?
>   * Who exactly is `env::current_account_id`, `env::predecessor_account_id`, `env::signer_account_id`, and how are these set?
> * How can a contract let user A send NEAR to user B? This is what I was trying to capture with the `send_funds` method, but it looks what I was trying to convey was nicely captured [in the docs you shared](https://www.near-sdk.io/promises/token-tx).
> * Understanding the on-chain storage components of a contract. I thought that when a contract was deployed, each time the contract was called, the `new` method would be called, and the properties in `new` would be set each time the contract was called.
>   
>   * On a similar note, seeing code like `LookupMap::new(b"my_property".to_vec())` was really confusing to me. It'd be nice if there was some additional clarification on the key / value paradigm for on-chain storage, outlined in the [overview here](https://docs.near.org/docs/concepts/data-storage#overview). Now I understand that `"my_property".to_vec()` sets the key for the storage object on-chain. At least I think I understand it :)
> 
> Let me know if this makes sense or not. Thanks!

RNHTTR:
I think it's okay to close. In my opinion, having some of this info up front would make the learning curve a little gentler. Just food for thought :)

Thanks for the explanations!

PR #833, opened 2021-08-30
Explained how to deal with empty method args for near cli

BenKurrek:
Added note saying: 

**Note:** There are two ways to deal with methods that require empty `{ args }`. Either send `{"field": null}` or simply omit this field and pass in nothing: `{}`

in the near-cli near call section.


thisisjoshford (approved):
👍🏼

PR #841, opened 2021-09-04
link to near-sdk.io site in the Rust SDK section

mikedotexe:
This effort is to provide a link to the SDK Docs site so that search engines can begin indexing this site. While it's in good shape, we cannot consider it fully complete at this time. When that time comes, we can discuss reworking other parts of this page in docs.


bucanero (approved):

bucanero:
LGTM 👍

PR #844, opened 2021-09-13
Fixed `#[payable]` example

maoueh:
It seems there is no import that exists for `payable` like the docs was implying. The macro simply needs to be put inside the wrapped `near_bindgen`.


austinabell (approved):
this is correct! I didn't know these docs existed

PR #847, opened 2021-09-16
Docusaurus v2 migration

bucanero:
This PR updates Docusaurus to v2


thisisjoshford (approved):
Made some minor CSS fixes / adjustments, but aside from that killer work @bucanero !

PR #866, opened 2021-09-28
Create tutorial: Minting NFTs using IPFS

bucanero:
Fixes https://github.com/near/devrel/issues/323


docs/tutorials/contracts/nfts/minting-nfts.md
```diff
@@ -0,0 +1,356 @@
...
+}
+```
+
+</p>
+</details>
+
+Now let's mint our first token! The following command will mint one copy of your NFT. Replace the `media` url with the one you [uploaded to IPFS](#uploading-the-image) earlier:
+
+```bash
+near call $ID nft_mint '{"token_id": "0", "receiver_id": "'$ID'", "token_metadata": { "title": "Some Art", "description": "My NFT media", "media": "https://bafkreiabag3ztnhe5pg7js4bj6sxuvkz3sdf76cjvcuqjoidvnfjz7vwrq.ipfs.dweb.link/", "copies": 1}}' --accountId $ID --deposit 10
```
thisisjoshford, line 266: Why do we attach 10N to this call?

docs/tutorials/contracts/nfts/minting-nfts.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+</p>
+</details>
+
+> **Note:** For `mainnet` you will need to prepend your command with `NEAR_ENV=mainnet`. [See here](/docs/tools/near-cli#network-selection) for more information.
+
+### Minting your NFTs
+
+The NFT contract should be initialized before usage, but for now we'll initialize with the default metadata:
```
thisisjoshford, line 230: Maybe we should speak on initializing contracts and why we do this. Also, include a tip that reminds people that they will need to delete their account if they have already deployed a contract.

thisisjoshford (commented):
Made some updated and left a couple of comments. :)

docs/tutorials/contracts/nfts/minting-nfts.md
bucanero: I was following the instructions from https://examples.near.org/NFT 
I assume that setting 10N is a way to avoid any issue due to insufficient gas, but I don't really know. 

Should we use a different value?

bucanero: Updated to 0.1 N

docs/tutorials/contracts/nfts/minting-nfts.md
bucanero: Added some notes about contract initialization. I kept the notes about re-using/deleting an account in the `Deploying the contract` section.

thisisjoshford (approved):

PR #880, opened 2021-10-14
Adding Hetzner pricing for NEAR sever cost estimation list

miohtama:
People are being ripped off with cloud server pricing, when a dedicated hardware server is much more cost-efficient for running a node. 

https://www.hetzner.com/dedicated-rootserver/matrix-ax


bucanero:
Node documentation has moved to https://github.com/near/node-docs

PR #898, opened 2021-11-03
Change pinned versions of docs.rs links to use "latest"

mikedotexe:
Self-explanatory, we want folks to see links to the latest version of the Rust docs.


bucanero (approved):
nice catch! Thanks for your help @mikedotexe 👍

PR #913, opened 2021-11-24
replace check mark

danielwpz:
the original x was ambiguous, check marks clearly indicate supporting of one action


thisisjoshford (approved):
Makes sense. :) Thanks @danielwpz !

PR #917, opened 2021-11-26
Reword the CPU requirements

nagisa:
Here are some motivating points for the changes here:

* 8-core (16-thread) implies that SMT is required to run `neard` which
  is not exactly the case. There are a couple modern intel chips which
  have SMT disabled, and a fair number of people running validators may
  want to disable SMT due to CPU vulnerability concerns. OpenBSD in
  particular disables SMT by default.
* i7/Xeon does not particularly mean anything much in terms of
  performance characteristics. A current generation Alder Lake i5 can
  easily outperform an i7 from the previous generation. Xeon is useful
  as an identifier for server-class hardware and features (such as ECC
  memory), but given that i7 is okay…
* We require a CPU with the x86_64-v2 feature level + AVX. “AVX support”
  does not really convey this adequately, so I went ahead and listed the
  specific CPU features we are looking for.

### DIFF
--- a/docs/develop/node/validator/hardware.md
+++ b/docs/develop/node/validator/hardware.md
@@ -12,23 +12,35 @@ This page covers the minimum and recommended hardware requirements for engaging 
 
 | Hardware       |  Recommended Specifications                                                  |
 | -------------- | ---------------------------------------------------------------              |
-| CPU            | 8-Core (16-Thread) Intel i7/Xeon or equivalent with AVX support              |
+| CPU            | x86_64 (Intel, AMD) processor with at least 8 physical cores                 |
+| CPU Features   | CMPXCHG16B, POPCNT, SSE4.2, AVX                                              |
 | RAM            | 16GB DDR4                                                                    |
 | Storage        | 250GB SSD (NVMe SSD is recommended. HDD will be enough for localnet only)    |
 
-_Verify AVX support on Linux by issuing the command ```$ lscpu | grep -oh  avx```. If the output is empty, your CPU is not supported._
+Verify CPU feature support by running the following command on Linux:
 
+```
+lscpu | grep -P '(?=.*avx )(?=.*sse4.2 )(?=.*cx16 )(?=.*popcnt )' > /dev/null \
+  && echo "Supported" \
+  || echo "Not supported"
+```
 
 ## Minimal Hardware Specifications {#minimal-hardware-specifications}
 
 | Hardware       |  Minimal Specifications                                                     |
 | -------------- | ---------------------------------------------------------------             |
-| CPU            | 8-Core (16-Thread) Intel i7/Xeon or equivalent with AVX support             |
+| CPU            | x86_64 (Intel, AMD) processor with at least 8 physical cores                |
+| CPU Features   | CMPXCHG16B, POPCNT, SSE4.2, AVX                                             |
 | RAM            | 8GB DDR4                                                                    |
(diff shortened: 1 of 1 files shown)


docs/develop/node/validator/hardware.md
```diff
@@ -12,23 +12,35 @@ This page covers the minimum and recommended hardware requirements for engaging
 
 | Hardware       |  Recommended Specifications                                                  |
 | -------------- | ---------------------------------------------------------------              |
-| CPU            | 8-Core (16-Thread) Intel i7/Xeon or equivalent with AVX support              |
+| CPU            | x86_64 (Intel, AMD) processor with at least 8 physical cores                 |
+| CPU Features   | CMPXCHG16B, POPCNT, SSE3, SSE4.1, SSE4.2, SSSE3, AVX                         |
```
matklad, line 16: Is there some nice "layman" formulation we can use to give a rough estimate of CPU performance? Ie, I think what `i7` tries to convey here is "reasonably fast, but still commodity" CPU. 

I wonder if `eg "<the exact CPU model we use for our validators>"` would work? 

>| CPU Features   | CMPXCHG16B, POPCNT, SSE3, SSE4.1, SSE4.2, SSSE3, AVX 

Curious, can the CPU support SSE4.2 but not SSE4.1? That is, can we transitively reduce this set?

bowenwang1996: Actually I think i7 may even be an overkill. For the nodes we run on gcp, the vcpu performance is definitely much worse than i7

nagisa: > Curious, can the CPU support SSE4.2 but not SSE4.1?

Yes. Most trivially in an virtualization context, where you can turn off and on whatever features you want. In practice I don't believe there is any hardware that will support SSE4.2 but not SSE4.1. Except maybe if there's some errata and the vendor disables the functionality entirely, but that seems unlikely to happen for SSE instruction sets.

I think it does indeed make sense to reduce the set here, just like I did for the test code below.

---

> I think what i7 tries to convey here is "reasonably fast, but still commodity" CPU.

I mentioned in the commit message the following reasoning:

> i7/Xeon does not particularly mean anything much in terms of
performance characteristics. A current generation Alder Lake i5 can
easily outperform an i7 from the previous generation. Xeon is useful
as an identifier for server-class hardware and features (such as ECC
memory), but given that i7 is okay…

To add to this, I think it'll be very difficult to find a low performance chip that runs `x86_64`, has 8+ cores, supports AVX and lacks the instruction throughput in some other way (e.g. is a low GHz part). It makes little economic sense to have a product line like that. (Unless we count Intel and some of their Xeons as terrible as [this one](https://ark.intel.com/content/www/us/en/ark/products/123540/intel-xeon-bronze-3106-processor-11m-cache-1-70-ghz.html); this part is literally worse than some Atoms you could get at the time)

bowenwang1996 (approved):

bucanero:
is this PR ready to be merged?

nagisa:
I'll reduce the set of the required features to just SSE4.2, give me a moment.

nagisa:
The CI failure does not seem related to the changes here. I don't know what's the right procedure to restart the test run, but think it is ready otherwise.

bucanero:
> The CI failure does not seem related to the changes here. I don't know what's the right procedure to restart the test run, but think it is ready otherwise.

thanks @nagisa , I'm merging the PR 👍

PR #922, opened 2021-12-01
Update gas.md

humanman:
removed line referencing "Did you Know?" section in /accounts page as it is no longer there

### DIFF
--- a/docs/concepts/gas.md
+++ b/docs/concepts/gas.md
@@ -222,7 +222,9 @@ In this sense, prepaid gas can be realized using a funded account and related co
 
 _So how can a developer pay the gas fee for their users on NEAR?_
 
-A user can use the funds directly from the developers account suitable only for the gas fees on this dApp. Then the developer has to distinguish users based on the signers' keys instead of the account names.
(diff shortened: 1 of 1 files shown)


docs/concepts/gas.md
```diff
@@ -224,8 +224,6 @@ _So how can a developer pay the gas fee for their users on NEAR?_
 
 A user can use the funds directly from the developers account suitable only for the gas fees on this dApp. Then the developer has to distinguish users based on the signers' keys instead of the account names.
```
bucanero, line 225:
```suggestion
- A user can use the funds directly from the developer's account suitable only for the gas fees on this dApp. Then the developer has to distinguish users based on the signers' keys instead of the account names.
```

docs/concepts/gas.md
```diff
@@ -224,8 +224,6 @@ _So how can a developer pay the gas fee for their users on NEAR?_
...
```
bucanero, line 226:
```suggestion

- Using function calls, you can allow a new user without an account to use your dApp and your contract on-chain. The back-end creates a new access key for the user on the contract's account and points it towards the contract itself. Now the user can immediately use the web app without going through any wallet.

```

bucanero (approved):

PR #923, opened 2021-12-07
Removed `keyStore setup` link under `Connect` section

Rishabhraghwendra18:
Removed `keyStore setup` link under `Connect` section because the link was redirecting to same page instead of any other page . 
I don't find any reason for redirecting user to same page , so I thought of removing it . It can be useful if the link redirects the user to some another page which may have a javascript file with whole config code written so that user can directly copy & paste it in it's code.


bucanero (commented):
The link is valid, and redirects to the `#key-store` anchor from a previous section.

PR #925, opened 2021-12-08
passed `.near-credentials` path in `UnencryptedFileSystemKeyStore()`

Rishabhraghwendra18:
passed `.near-credentials` path in `UnencryptedFileSystemKeyStore()` which would otherwise give error:
`The "path" argument must be of type string. Received undefined`

It will be good idea to tell user to pass `.near-credentials` file path otherwise he/she will be confused about it and has to search a level more about it in docs .
I get the idea about it from this [page](https://docs.near.org/docs/api/naj-quick-reference#key-store)

### DIFF
--- a/docs/develop/front-end/introduction.md
+++ b/docs/develop/front-end/introduction.md
@@ -240,11 +240,16 @@ npm install near-api-js
 To do anything useful on the NEAR platform you first have to establish a connection.
 
 ```js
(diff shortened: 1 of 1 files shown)


doriancrutcher (approved):
This does omit the error I also get when running a NAJ node JS file. I approve these changes

PR #926, opened 2021-12-08
`path` is undefined without requiring `path` module

Rishabhraghwendra18:
`path` is undefined without requiring (importing) `path` module in node.js


doriancrutcher (approved):
I agree with this change. Every time I setup a NAJ  nodeJS file I always have to add that line.

Rishabhraghwendra18:
@doriancrutcher sorry to ask but what is NAJ ?

PR #930, opened 2021-12-14
Instructions for nodes upgrading alert based on public telemetry data

nikurt:


chefsale (approved):

kucharskim:
I'm not Grafana person, so cannot judge the queries, but this source of information is definitely very useful, in the time being!

docs/develop/node/maintenance/upgrade_alert.md
```diff
@@ -0,0 +1,156 @@
...
+
+Validator nodes periodically submit telemetry data, which is stored in a publicly-accessible PostgresSQL database.
+Note that the nodes can opt out of submitting the telemetry data, and the telemetry data isn't guaranteed to be accurate.
+
+With these caveats clarified, let's define an upgrade alert. The instructions were tested only in Grafana.
+
+Step 1. Add a PostgreSQL data source with the following credentials:
+
+* [Testnet](https://github.com/near/near-explorer/blob/master/backend/config/env-indexer-testnet#L14-L17)
+* [Mainnet](https://github.com/near/near-explorer/blob/master/backend/config/env-indexer-mainnet#L14-L17)
```
kucharskim, line 24: I think lines don't match. It points to `indexer_analytics_mainnet` database but based on tables it should be `telemetry_mainnet`?

kucharskim: I would skip lines in URL and just name which database is source of truth. It's more likely that lines can go off when file is modified, than the database name will change.

kucharskim: ..also grepping a repo or multiple repositories for a string (a database name) is easier than looking for specific line references. In case database name indeed will change.

nikurt: Thanks, updated.

PR #952, opened 2022-01-15
Removes CreateTodo and TodoList from initial App.js setup

elliotBraem:
These components have not been made yet in the tutorial, so npm run start fails


bucanero:
thanks @elliotBraem 👍

PR #958, opened 2022-01-18
Kurtosis dApp Config

BenKurrek:
Fixes https://github.com/near/devrel/issues/450
Changed dApp config in kurtosis tutorial to use env variables rather than copy paste


mieubrisse (approved):
Should work, but asked some folks on the #kurtosis channel to verify since my guestbook dApp has gone back to not working :(

thisisjoshford (approved):

PR #960, opened 2022-01-19
Add light mode

janewang:


frol (approved):
We should merge it as is, but it will be great to have a follow-up PR with a description of what is the point of the light node.

PR #978, opened 2022-02-01
running-a-node-macos-linux: improvements to backup instructions

mina86:
First of all, `neard` does not create an empty `data` directory when
initialising a home dir.  It has to be created manually.  This can be
easily done with `mkdir` of course.

Second of all, avoid changing working directory by using tar’s `-C`
flag.  This is especially meaningful with `neard` instructions since
it uses `./neard` to launch the node so the command is sensitive to
working directory.

Lastly, `wget -O - …` will not produce an output file therefore there
is no reason to `rm data.tar`.


bucanero:
Hi @janewang, could you review and merge these node-related changes? thank you 🙏

janewang (approved):

PR #995, opened 2022-02-11
Improve data snapshot instructions

mina86:
Firstly, `wget <file> <url>` is an invalid invocation so fix that to
using `-P` option.  (Another alternative would be to use `-O` option
but since existing documents already use `-P` use that consistently in
Node Data Snapshots section as well).  With that, move `-P` at the
front of the command so that it’s more obvious that the switch is
there and is necessary.

Secondly, for better or worse (*cough* worse *cough*) the `data.tar`
files do not include the `data` directory so make sure that
instructions indicate that the file needs to be extracted inside of
the `~/.near/data` directory.

### DIFF
--- a/docs/develop/node/archival/run-archival-node-without-nearup.md
+++ b/docs/develop/node/archival/run-archival-node-without-nearup.md
@@ -128,7 +128,7 @@ or
 
 ```bash
 $ rm ~/.near/config.json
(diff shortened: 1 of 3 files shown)


mina86:
OK, I give up. I spent couple of hours trying to make sense of the documentation.  There’s dozen of places where the same information is repeated and updating the documentation is a just a nightmare.  Trying to make it DRY is even worse.

PR #1007, opened 2022-02-21
Add information to CLI delete call description

idea404:
Question was asked in #dev-support in Discord relating to this matter. 
I propose adding this text here to address this concern in our documentation going forward.

### DIFF
--- a/docs/tools/near-cli.md
+++ b/docs/tools/near-cli.md
@@ -604,7 +604,9 @@ near send sender.testnet receiver.testnet 10
 
 > Deletes an account and transfers remaining balance to a beneficiary account.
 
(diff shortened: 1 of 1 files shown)


docs/tools/near-cli.md
```diff
@@ -604,6 +604,8 @@ near send sender.testnet receiver.testnet 10
 
 > Deletes an account and transfers remaining balance to a beneficiary account.
 
+**Note:** The deleted account's name will be made available for use after this action is performed.
```
bucanero, line 607:
```suggestion
:::note
The deleted account's name will be made available for use after this action is performed.
:::
```

bucanero (approved):

Issue #1018, opened 2022-02-28
Cannot start docusaurus


mikedotexe:
I heard feedback from some developers that were unable to contribute to the docs. Here is their screenshot:

![image](https://user-images.githubusercontent.com/1042667/156070025-3763444a-4fbd-4928-a417-70714e0d02d5.png)

I pulled `master` and was able to reproduce this problem.

thisisjoshford:
Works fine locally for me but was able to reproduce in a VM... Working on a fix now.

gagdiez:
@thisisjoshford Having exactly the same problem, I am working on solving it. If you have any ideas please let me know

gagdiez:
Managed to make it compile, but the search bar is not working locally

PR #1019, opened 2022-03-02
Fixed compiling errors

gagdiez:
I managed to make the documentation compile, however, I was not able to make the search bar work locally.

This fixes issue #1018


website/package.json
```diff
@@ -18,19 +18,20 @@
...
     "@saucelabs/theme-github-codeblock": "^0.1.1",
     "clsx": "^1.1.1",
-    "react": "^17.0.1",
-    "react-dom": "^17.0.1"
+    "docusaurus-theme-search-typesense": "^0.3.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2"
   }
-}
+}
```
bucanero, line 37:
```suggestion
}

```

bucanero:
@gagdiez thanks for the update

btw, have you tried running `yarn build` locally? I'm testing on my end and I'm getting this error:

```
[typesense-instantsearch-adapter] Please use snake_cased versions of parameters in additionalSearchParameters instead of camelCased parameters. For example: Use query_by instead of queryBy. camelCased parameters will be deprecated in a future version. We're making this change so that parameter names are identical to the ones sent to Typesense (which are all snake_cased), and to also keep the types for these parameters in sync with the types defined in typesense-js.


TypeError: chalk_1.default.bold is not a function
```

I don't want to merge before being sure that this builds our docs properly. For some reason, this PR is not running the Render preview or build checks. 🤔

gagdiez:
@bucanero I have a different error

```
❌ Configuration file is invalid. Check the following parameters in your configuration file:
        - Required option 'api_token' is missing
```

PR #1064, opened 2022-05-13
Pessimitic gas prices

jakmeier:


jakmeier:
Hey, I noticed a while ago that pessimistic gas prices aren't really documented anywhere. So I thought I write things down and try to fit it in somewhere in the docs.

@gagdiez Am I submitting this to the right place keeping the docs reorganisation in mind? I really like the idea of assigning clear target audiences for different documentation parts. With that in mind, I thought this is something that probably is mostly relevant for tech enthusiasts, thus I guess "Concepts" is the right place. (Devs don't need to understand this at all, everything happens magically in the background. But they might stumble over it when they look at gas number in the explorer.)

gagdiez:
Yes, this is the right place! Thank you for the contribution, we actually get a lot of questions about GAS pricing, and this really helps to clarify it.

gagdiez (approved):

jakmeier:
Cool! I was quite confused about gas prices myself only a few months ago. I had to go through the source code in many instances to understand what *actually* happens.

I would be happy to help improve documentation around gas further. Maybe if you know of specific questions that come up often, I could look into writing something about it.

PR #1065, opened 2022-05-13
Update example deployment cost

jakmeier:
With protocol version 53 (release 1.26) the deployment gas cost per byte
increases from 6_812_999 to 64_572_944.

This updates the example gas calculation.


jakmeier:
See also the change PR on nearcore: https://github.com/near/nearcore/pull/6397

This change reaches mainnet next week. Thus, it is already observable on testnet. So I think updating the docs right about now wouldn't hurt.

bucanero (approved):

PR #1076, opened 2022-06-03
contribute-nearcore: add info on ROCKSDB_LIB_DIR and other minor updates

mina86:
Most notably, add a section about compiling while using rocksdb
library provided by the system.  Based on my observation, this helps
with OOMs.  I think (though I haven’t tested it) that rust-rocksdb
builds rocksdb library with LTO enabled which consumes bunch of
memory.

While at it, make some minor fixes to the document.  Change ‘bash’
preformatted blocks to ‘console’ where the latter is more appropriate
and remove mention of `start_localnet.py` script which no longer
exists.

### DIFF
--- a/docs/community/contribute/nearcore.md
+++ b/docs/community/contribute/nearcore.md
@@ -38,8 +38,8 @@ cd nearcore
 Navigate to the root of the repository, and run:
 
 ```bash
-cargo run --package neard --bin neard -- init
-cargo run --package neard --bin neard -- run
+cargo run -p neard -- init
+cargo run -p neard -- run
 ```
 
 This will setup a local chain with `init` and will run the node.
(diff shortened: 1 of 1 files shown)


docs/community/contribute/nearcore.md
```diff
@@ -149,14 +150,14 @@ lto=off`.
 
 Requires installing lld linker.
 
-```bash
-# cargo clean
-# time RUSTFLAGS='-C lto=off -C link-arg=-fuse-ld=lld' RUSTC_WRAPPER= cargo build -p neard --release
+```console
+$ cargo clean
+$ time RUSTFLAGS='-C lto=off -C link-arg=-fuse-ld=lld' RUSTC_WRAPPER= cargo build -p neard --release
```
nagisa, line 155: NB: this is going to be shortly made more straightforward via [`-Clinker-flavor=gcc:lld`](https://github.com/rust-lang/rust/pull/96827), with or without `-Clink-self-contained=linker`. We should keep an action item for that.

docs/community/contribute/nearcore.md
```diff
@@ -166,32 +167,69 @@ sys     0m4.307s
...
+By default, building `neard` will result in compiling `librocksdb`
+which takes up some non-negligible amount of memory especially when
+building on a system with many cores.  It’s possible to configure
+build process to provide system-provided library instead.  This is
+done by setting `ROCKSDB_LIB_DIR` environment variable to location
+where `librocksdb.a` file is installed:
+
+```console
+$ cargo clean
+$ time ROCKSDB_LIB_DIR=/usr/lib/x86_64-linux-gnu cargo build -p neard --release
```
nagisa, line 194: nit: using the `env` command (`env VAR=val VAR2=val2 command...`) makes supplying these environment variables in this style less dependent on the specific shell used.

nagisa:
largely LGTM.

PR #1077, opened 2022-06-03
Mieubrisse/more complete neartosis docs

mieubrisse:


mieubrisse:
@thisisjoshford bunch of quality-of-life improvements for users

docs/tools/kurtosis-localnet.md
```diff
@@ -203,23 +217,17 @@ The Kurtosis Team has created a great [video presentation](https://www.loom.com/
...
-```bash
-http://127.0.0.1:51395
-```
+The account creation is exactly the same as on mainnet or testnet but **only the passphrase recovery mode** will work here. Also note that the root account is `test.near` instead of `testnet` or `mainnet`. This means that all the accounts you create will be [subaccounts](/docs/concepts/account#subaccounts) of `test.near`. (ex. `benji.test.near`)
 
-- Click on the URL displayed in the terminal to launch your local NEAR Wallet or copy/paste this address in your browser.
+:::tip
 
-![Local wallet landing page](/docs/assets/kurtosis/local-wallet-landing-page.png)
+Because the Wallet will start on the same same host and port, and the Wallet stores your account information in your browser's local storage, if you've created an account with a previous Kurtosis network then the Wallet will be storing the old account's information (which won't work with your new network). If this is the case for you, you'll need to clear the Wallet's storage by visiting [your Chrome's local storage](chrome://settings/siteData?searchSubpage=127.0.0.1), removing the `127.0.0.1` entry, and refreshing the Wallet.
```
bucanero, line 228:
```suggestion
Because the Wallet will start on the same host and port, and the Wallet stores your account information in your browser's local storage, if you've created an account with a previous Kurtosis network then the Wallet will be storing the old account's information (which won't work with your new network). If this is the case for you, you'll need to clear the Wallet's storage by visiting [your Chrome's local storage](chrome://settings/siteData?searchSubpage=127.0.0.1), removing the `127.0.0.1` entry, and refreshing the Wallet.
```

bucanero (approved):

PR #1093, opened 2022-06-22
Add warning for the delete command

SoftwareAndOutsourcing:


SoftwareAndOutsourcing:
Use case:

1. A user request a wallet address in a CeFi market.
2. The user wants to send all funds to the CeFi wallet, it uses `delete` instead of `transfer`.

Result:

As the CeFi wallet is not initialized, the funds are lost.

Expected:

The `delete` action should initialize the wallet so the transfer is completed successfully.

This use case could also indicate a `delete` command design flaw.

gagdiez:
We added the warning on the new documentation that is now merged. Thanks for reporting it!

PR #1096, opened 2022-06-28
Add banner to introduce new-docs

bucanero:


gagdiez:
Can we change it to: "Did you know we are updating our docs? Check the preview and let us know what you think!". And that is a whole link.

PR #1098, opened 2022-07-04
Updates example link for attaching tokens

dimosmera:
URL is out of date

Currently used: https://github.com/near-examples/guest-book/blob/master/src/App.js#L27-L30

New one: https://github.com/near-examples/guest-book-tutorial/blob/main/src/App.js#L32-L33


bucanero (approved):

bucanero:
thanks @dimosmera !

PR #1106, opened 2022-07-18
Add Smart Contract Section

idea404:
This PR adds content for the smart contract and state sections in `1.basics` under `1.concepts`


idea404:
@gagdiez @bucanero I added these paragraphs to this section, by adding more I felt it would overlap too much with other parts of the docs. In part I'm not sure if this info is really adding anything here. What do you think? 

Guillermo did you have something like this in mind here?

gagdiez:
@idea404 this section is more "explaining what a smart contract is to your grandma", so we only really need to explain that a smart contract is a small program stored on a blockchain, that anyone can access it, and that it has a storage (that anyone can see but only the contract change).

idea404:
I've simplified the contents for both sections @gagdiez. What do you think?

PR #1116, opened 2022-07-29
Added Wallet Selector in sidebar with all wallet docs

amirsaran3:
Added "Wallet Selector" category to "Accounts & Transactions" sub-section on "Tools" page. Category has docs for each currently available wallet.


website/docusaurus.config.js
```diff
@@ -62,9 +62,9 @@ module.exports = {
           anonymizeIP: true,
         },
         blog: {},
-        theme: {
-          customCss: "/src/css/customTheme.css",
-        },
+        // theme: {
+        //   customCss: "/src/css/customTheme.css",
+        // },
```
bucanero, line 67: why are you disabling the CSS theme?

bucanero (changes requested):
please restore the CSS theme

website/docusaurus.config.js
bucanero, line 67:
```suggestion
         theme: {
           customCss: "/src/css/customTheme.css",
         },
```

gagdiez:
@amirsaran3 this looks like low level (API reference) documentation. I would strongly recommend to put it on its own website/github pages.

In our docs we will only want a high level explanation on how to use the wallet selector.

website/docusaurus.config.js
amirsaran3: Hey. Sorry for not responding. I had problems running the development server while this theme was here. So I removed it just to be able to work. I pushed new changes that does not remove it.

PR #1126, opened 2022-08-09
Update naj-wallet.md

flmel:
https://near-docs.io/tools/near-api-js/wallet#wallet-connection
has been changed to use tabs but they have not been imported


bucanero:
good catch! 👍

PR #1174, opened 2022-08-26
clarifying this 03-intermediate/02-use-seed-phrase.md lesson

ryancwalsh:
See original PR at https://github.com/near/sdk-docs/pull/119

This part of the tutorial took me longer to understand than it probably needed to.

My hope is that my proposed edits will help others understand it quicker and easier.

Please carefully check that I'm right (especially at the part where I say "The original puzzle creator would have already called this same function with the same inputs and would have already called AddKey on this contract...").

### DIFF
--- a/docs/3.tutorials/crosswords/03-intermediate/02-use-seed-phrase.md
+++ b/docs/3.tutorials/crosswords/03-intermediate/02-use-seed-phrase.md
@@ -6,12 +6,12 @@ title: "Implementing the seed phrase logic from the necessary libraries"
 
 # Seed phrase and key derivation
 
-There are two things we'll want to do:
+There are two separate things we'll want to do:
 
-1. **Create a random seed phrase** for the user when they visit the crossword puzzle. This will be used if they win and don't have a NEAR account and wish to create one. They can then paste this seed phrase into NEAR Wallet afterward.
+1. **Create a random seed phrase** for the user when they visit the crossword puzzle. This will be used if they win and don't have a NEAR account and wish to create one. They can then paste this seed phrase into NEAR Wallet afterward to import their account (which is basically like "logging in" and is currently possible at https://wallet.near.org/recover-seed-phrase).
 2. **Turn the crossword solution into a key pair**, instead of just hashing it.
 
-## Seed phrase
+## near-seed-phrase library
 
 We can add the `near-seed-phrase` package to our project with:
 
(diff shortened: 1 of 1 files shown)


docs/3.tutorials/crosswords/03-intermediate/02-use-seed-phrase.md
```diff
@@ -23,22 +23,27 @@ At this point in the tutorial, it's more difficult to share code snippets that a
...
+(This security measure prevents front-running.)
 
 ```js
 import { parseSeedPhrase } from 'near-seed-phrase';
-// Get the seed phrase from the completed puzzle
-let seedPhrase = parseSolutionSeedPhrase(data, gridData);
+// Get the seed phrase from the completed puzzle. 
+// The original puzzle creator would have already called this same function with the same inputs and would have 
+// already called `AddKey` on this contract to add the key related to this seed phrase. Here, using this deterministic 
+// function, the front-end will automatically generate that same key based on the inputs from the winner.
```
idea404, line 45: this isn't really correct. What happens is that the result for the puzzle is encrypted and the public key for that is saved on the smart contract. If this puzzle has the right outcome then its private key should be able to decrypt the information saved on the smart contract and is thus the right solution

idea404 (changes requested):
First half looks good, the second half on puzzle encryption could be left out

docs/3.tutorials/crosswords/03-intermediate/02-use-seed-phrase.md
ryancwalsh: @idea404 Thanks. I think I'm still having trouble understanding. I'll walk through how I'm thinking here.

At https://github.com/near-examples/crossword-tutorial-chapter-3/blob/969e25b1cc85cbcfe98071972607428789f90d3a/contract/src/lib.rs#L285 I see that the puzzle creator publishes a public key that is derived from the seed phrase that was derived from the official crossword answer.

(I also see `answers` as a parameter, even though I thought we didn't want to publicly store the correct answers, so I was surprised by this and must be confused.)

Then, see `add_access_key` at https://github.com/near-examples/crossword-tutorial-chapter-3/blob/969e25b1cc85cbcfe98071972607428789f90d3a/contract/src/lib.rs#L315 This is what I meant when I said _The original puzzle creator would have called `AddKey`_.

For the contestant proposing an answer, I see https://github.com/near-examples/crossword-tutorial-chapter-3/blob/969e25b1cc85cbcfe98071972607428789f90d3a/src/App.js#L146 which similarly derives a seed phrase from the possible solution and then from that seed phrase derives a private key.

Can you please help clear up. what I'm misunderstanding? I think this conversation we're having will be valuable for the docs. 

In particular I was confused by `the result for the puzzle is encrypted` and `private key should be able to decrypt the information saved on the smart contract`.  I usually think of the words "encrypt/decrypt" as only applying in bidirectional cases with keys (not in cases like `parseSolutionSeedPhrase`, which seems more like a one-way hash / like a https://en.wikipedia.org/wiki/PBKDF2).

Thanks! 😃

idea404: The answers aren't publicly stored, but their encrypted public keys are. Once encrypted, they become gibberish which can only be reverted back to the legible answer by the private key that is generated by "parsing" the words from the correct solution (that is, creating a ed25519 key-pair from the puzzle's solution words). The creation of the puzzle is done by the dev of the smart contract apparently, and saved in the smart contract as explained here: https://github.com/near-examples/crossword-tutorial-chapter-3/blob/969e25b1cc85cbcfe98071972607428789f90d3a/contract/src/lib.rs#L285

The puzzle solution private key, generated by converting the solution words to a key-pair by the frontend, is used to sign a transaction you can see here: https://github.com/near-examples/crossword-tutorial-chapter-3/blob/969e25b1cc85cbcfe98071972607428789f90d3a/src/App.js#L154

ryancwalsh: @idea404 Thanks for your response. I cloned the repo locally and started going through it line by line. I found the Readme and tried to follow the steps. I've submitted a PR at https://github.com/near-examples/crossword-tutorial-chapter-3/pull/4/files

I still haven't been able to understand what about my explanation was incorrect. We'll work through it, though. Once we have a repo whose steps work, it will be helpful for everyone. Thanks.

idea404: @ryancwalsh On second thought I believe I misunderstood your comments in the code above. I thought you meant the puzzle's solution was saved as a key without the encryption step which makes it into a pub key first. Taking into consideration this is not the case, we can leave this in

idea404 (approved):
LGTM

docs/3.tutorials/crosswords/03-intermediate/02-use-seed-phrase.md
ryancwalsh: @idea404 Ok, thanks. If I have more ideas for clarification or questions as I learn more, I'll be sure to send. Thanks!

PR #1185, opened 2022-09-07
Support versioning for packages

maxhr:
- Use docs-generator for including TypeDocs of near-api-js and near-sdk-js in this repo


bucanero:
it seems that all the other Github actions that we had, (link checker, build check, spell checker) didn't run in this PR. 🤔 

The Render preview wasn't generated either. I wonder what happened

PR #1203, opened 2022-09-13
Google search integration

maxhr:
- Google search
- a friendlier 404 page


gagdiez:
I have mixed feelings about this feature, because we lose the real-time search, and introduces a bunch of **adds** in the result.

thisisjoshford:
closing this PR as we've decided to stick with Algolia for now :)

PR #1208, opened 2022-09-16
Update introduction.md

nagendraK6:
Fixing a typo in the documentation. 
Original content
Near SDK comes in three flavors

Expected content
Near SDK comes in two flavors


gagdiez:
Interesting, actually, the error is that it should say "Rust, JS, and Assemblyscript"

nagendraK6:
I will send another pull requests

PR #1215, opened 2022-09-20
JS SDK Nesting Example

idea404:
New area to docs: 

image

### DIFF
--- a/docs/2.develop/contracts/storage.md
+++ b/docs/2.develop/contracts/storage.md
@@ -100,7 +100,8 @@ Implements a [map/dictionary](https://en.wikipedia.org/wiki/Associative_array) w
   </Language>
 </CodeTabs>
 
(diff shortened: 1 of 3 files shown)


idea404:
@bucanero I tried to fix the broken CI build check but can not seem to find the issue. The build works fine locally and on render

gagdiez:
@idea404 should we add this as a "detail" (that you need to open to see) since it will be gone soon?

idea404:
> @idea404 should we add this as a "detail" (that you need to open to see) since it will be gone soon?
> 
> ```
> <details>
> <summary>Title goes here</summary>
> </details>
> ```

Sounds good to me, will update

idea404:
@gagdiez 
image

gagdiez (approved):

PR #1221, opened 2022-09-25
Update count-near.md

Devendra116:
Updated image references


bucanero:
The image links are working, you can check the published version: https://docs.near.org/tutorials/examples/count-near

PR #1227, opened 2022-09-28
fix: gas is burned, not given to validators

jakmeier:
Currently this is wrong and in conflict with other parts of the documentation.

Gas really is just burned.

### DIFF
--- a/docs/1.concepts/basics/transactions/gas.md
+++ b/docs/1.concepts/basics/transactions/gas.md
@@ -33,7 +33,7 @@ In addition, NEAR implements two unique features with respect to gas fees:
 2. Allowing for **free** transactions
 
 ### 1. Gas as a Developer Incentive
-Something unique to NEAR is that the GAS is not used to pay validators. In transactions that involve calling a contract the fee is actually divided:
+Something unique to NEAR is that the GAS is not used to pay validators. In transactions where calling a contract would incur a gas fee, the fee is actually divided as follows:
 - 30% goes to the smart contract.
 - 70% is burned.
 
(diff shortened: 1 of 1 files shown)


docs/1.concepts/basics/transactions/gas.md
```diff
@@ -25,16 +26,16 @@ When you send a transaction to the NEAR network different [validators](../valida
...
+As many other networks, NEAR **pays the validators** for their job. Also like many other networks, users have to pay a small **fee** (aka **gas fee**) on every transaction. But instead of giving the gas fee to the validators, validators receive their reward independent from the gas fees. This topic is discussed in more details in the [validators](../validators.md) section.
 
-However, NEAR implements two unique features with respect to gas fees:
+In addition, NEAR implements two unique features with respect to gas fees:
 1. Sharing fees with developers
 2. Allowing for **free** transactions
 
 ### 1. Gas as a Developer Incentive
-Something unique to NEAR is that the GAS is not only used to pay validators. In transactions that involve calling a contract the fee is actually divided:
+Something unique to NEAR is that the GAS is not used to pay validators. In transactions that involve calling a contract the fee is actually divided:
```
idea404, line 36: Small suggestion for a little more clarity if you agree.

```suggestion
Something unique to NEAR is that the GAS is not used to pay validators. In transactions where calling a contract would incur a gas fee, the fee is actually divided as follows:
```

idea404 (approved):
Brilliant PR, thank you Jakob. Important nuance that also demonstrates the possible deflationary nature of the network once a given throughput is achieved. 

I wonder where else we'd have to make small adjustments to accommodate this, but I'll keep an eye out for it from now on.

jakmeier:
> I wonder where else we'd have to make small adjustments to accommodate this, but I'll keep an eye out for it from now on.

As far as I could see, this page was the only page in these docs where we explicitly write "pay gas to validators" or some variants of that. Specifically, the page about validators is already correct.

But yeah let's keep an eye out for this, we should be consistent about this across the board. Even seasoned members of the community, in my experience, often assume gas is paid to validators. Let's work towards squashing that misunderstanding as much as possible.

PR #1239, opened 2022-10-05
Mention ECDSA Public Key Recovery

idea404:
This PR adds a short section on recovering public keys from signed messages with the Rust SDK. 

It follows questions from Discord and is in line with this reply: https://discord.com/channels/490367152054992913/855524499755499520/958736044935766107

### DIFF
--- a/docs/2.develop/contracts/environment/environment.md
+++ b/docs/2.develop/contracts/environment/environment.md
@@ -147,4 +147,4 @@ If you already [estimated the Gas](/concepts/basics/transactions/gas#accurate-es
 
 ## Recover ECDSA Public Key from Signed Messages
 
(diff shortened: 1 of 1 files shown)


idea404:
would be good to have some input from Austin before this gets merged

bucanero (approved):

gagdiez:
This looks more appropriated for the NEAR-SDK-RS specific docs, since it only has the rust example. Otherwise, we should add the JS version as well.

idea404:
> This looks more appropriated for the NEAR-SDK-RS specific docs, since it only has the rust example. Otherwise, we should add the JS version as well.

Agreed. There is also a JS SDK version, will update this with the JS version as well

gagdiez:
I have an idea, what if we create the page "Environmental Functions" and document them all? apparently there are a couple of methods, and it would be super useful. @idea404

idea404:
> I have an idea, what if we create the page "Environmental Functions" and document them all? apparently there are a couple of methods, and it would be super useful. @idea404

Good idea. I'll close this PR and make a new one that includes this information among that for the other functions.

Issue #345, opened 2020-05-19
[BUG] Docs still talk about state rent


ilblackdragon:
**Describe the bug**
This page https://docs.near.org/docs/concepts/gas still has "state rent".

**Expected behavior**
Should be:
 - Called differently, as it's not only about gas 
 - Describe storage pricing (details are in "Storage" section of https://near.org/blog/near-protocol-economics/)

**Screenshots**
![photo_2020-05-18 17 28 38](https://user-images.githubusercontent.com/175486/82271583-08e49e80-992d-11ea-8a9e-100be4f138e0.jpeg)

**:pencil: Location**
https://docs.near.org/docs/concepts/gas

amgando:
#373 was merged and closes this

firatNEAR:
I am re-opening this issue since there are multiple places in Docs which still state `state rent`
* https://docs.near.org/integrator/faq#is-there-a-minimum-account-balance
* https://nomicon.io/RuntimeSpec/Scenarios/FinancialTransaction#processing-receipt
* https://near.org/papers/economics-in-sharded-blockchain/#transaction-and-storage-fees (Not sure if this document relates to this near/docs

thisisjoshford:
#1245 Fixes this issue

PR #1245, opened 2022-10-10
Fix rent model to storage deposit

idea404:
This PR addresses the docs part of this issue: https://github.com/near/docs/issues/345

### DIFF
--- a/docs/6.integrator/faq.md
+++ b/docs/6.integrator/faq.md
@@ -213,7 +213,7 @@ Users create accounts with human-readable names (eg `alice`) which can contain m
 
 ### Is there a minimum account balance? {#is-there-a-minimum-account-balance}
 
-To limit on-chain "dust", accounts (and contracts) are charged rent for storing data on the chain. This means that if the balance of the account goes below some `threshold * rent_on_block` then account can be removed by anyone. Also any user can remove their own account and transfer left over balance to another (beneficiary) account.
+To limit on-chain "dust", accounts (and contracts) are charged a refundable deposit for storing data on the chain. This means that if the balance of the account does not have enough balance to cover an increased deposit for additional storage of data, storing additional data will fail. Also, any user can remove their own account and transfer left over balance to another (beneficiary) account.
 
 There will be a restoration mechanism for accounts removed (or slept) in this way implemented in the future.
 


docs/6.integrator/faq.md
```diff
@@ -213,7 +213,7 @@ Users create accounts with human-readable names (eg `alice`) which can contain m
 
 ### Is there a minimum account balance? {#is-there-a-minimum-account-balance}
 
-To limit on-chain "dust", accounts (and contracts) are charged rent for storing data on the chain. This means that if the balance of the account goes below some `threshold * rent_on_block` then account can be removed by anyone. Also any user can remove their own account and transfer left over balance to another (beneficiary) account.
+To limit on-chain "dust", accounts (and contracts) are charged a refundable deposit for storing data on the chain. This means that if the balance of the account does not have enough balance to cover an increased deposit for additional storage of data, storing additional data will fail. Also, any user can remove their own account and transfer left over balance to another (beneficiary) account.
 
 There will be a restoration mechanism for accounts removed (or slept) in this way implemented in the future.
```
firatNEAR, line 218: I'm not sure if we should leave this here also, causes confusion. We should add the relevant details if/when this mechanism is implemented.

idea404: I'm confused. Errors are thrown in https://github.com/near/nearcore/blob/279044f09a7e6e5e3f26db4898af3655dae6eda6/runtime/runtime/src/lib.rs#L451-L739 and https://github.com/near/nearcore/blob/83b1f80313ec982a6e16a1941d07701e46b7fc35/runtime/runtime/src/verifier.rs#L75-L222 whenever there is not enough balance for storage. What do you mean?

firatNEAR: I meant this part sorry
``` There will be a restoration mechanism for accounts removed (or slept) in this way implemented in the future. ```

idea404: Ah great! Will remove then 👍

idea404:
@firatNEAR wdyt? 
@gagdiez @bucanero you guys OK with this too?

bucanero (approved):
LGTM

PR #1258, opened 2022-10-14
Update collections.md

zlog-in:
replacing for better understanding


gagdiez:
@zlog-in thanks for the contribution, however, I find the text more accurate and simpler to read that the new suggestion, for which I will close this PR.

PR #1260, opened 2022-10-18
Docs staging test

bucanero:
https://staging.near-docs.io/


bucanero:
@maxhr is this staging site still needed? Since we already merged Algolia's new search, let me know if we can close this PR and remove the staging site.

PR #1264, opened 2022-10-20
Three New Examples

gagdiez:
This PR adds 3 new examples:
- Complex Cross-Contract Calls
- Contract factory
- Self-upgrade & state migration


idea404 (approved):
LGTM. 

Would like to see JS SDK snippets of the code too, but we can address this on a later PR. 

Was not aware that running `pub fn migrate()` would overwrite state and that would just work, I always imagined this would cause conflicts of some sort. Does `env::state_read()` return all data collections from the contract? Would putting all that into one variable also not cause problems if the old contract has multiple data collections too? 

Really like the gas nuance with serialization/deserialization and passing raw input to the function too.

gagdiez:
@idea404 added some more text based on your comment. `migrate` works because it has the decorator `#[init(ignore_state)]`. `init` means that whatever it returns will be written as the state (overwriting the current one), `ignore_state` enables to ignore the fact that there is currently a state that's not compatible with the `Contract` struct.

PR #1272, opened 2022-10-25
Near wallet selector docs

DamirSQA:
Near wallet selector docs


gagdiez:
I think wallet selector goes in the same section as MEAR CLI, NEAR SDK, NEAR API JS, etc. Since it's a devs tool.

DamirSQA:
Thanks for suggestion, it's updated. You can check it out when you have a chance.

gagdiez (approved):

PR #1281, opened 2022-10-29
fix: cross contract call snippet used wrong code

luuuu-learning:
In the rust code snippet of [sending information](https://docs.near.org/develop/contracts/crosscontract#snippet-sending-information) from one contract to another, doc selected wrong code block from github. It should be `L51-L75` instead to show `change_greeting` and `change_greeting_callback`.

Current doc shows `query_greeting_callback` and `change_greeting`.


bucanero (approved):

bucanero:
thanks for the fix 👍

PR #1289, opened 2022-11-14
Add indexing services

bucanero:


website/sidebars.json
```diff
@@ -169,6 +169,7 @@
       "Monitor User Activity": [
         "tools/realtime",
         "tools/indexer-for-explorer",
+        "tools/indexing",
```
gagdiez, line 172: maybe it would be better to put this in the `Community Tools` section?

gagdiez (changes requested):

website/sidebars.json
bucanero: yes, you're right, Community Tools are a better fit for this kind of indexing tool list. 👍 I'll update the draft

bucanero: fixed

PR #1301, opened 2022-11-22
Introduce JS SDK Docs Section 

idea404:
JS SDK version of the rust-sdk.io port to docs

### DIFF
--- a/docs/sdk/js/contract-interface/private-methods.md
+++ b/docs/sdk/js/contract-interface/private-methods.md
@@ -47,7 +47,7 @@ helperMethod(a, b) {
 2. Using an internal helper function in the module scope.
 
 ```javascript
-// Function that can be called in another Rust file
+// Function that can be called in another JS file
 function getFirstName(account) {
   // ...
 }
--- a/docs/sdk/js/contract-interface/private-methods.md
+++ b/docs/sdk/js/contract-interface/private-methods.md
@@ -1,5 +1,71 @@
 ---
-sidebar_position: 3
+sidebar_position: 2
+title: "Private Methods"
 ---
 
-<!-- TODO -->
+# Private Methods
+
+## When Using Callbacks
+
+Usually, when a contract has to have a callback for a remote cross-contract call, this callback method should only be called by the contract itself to avoid someone else calling it and changing the state. A common pattern is to have an assertion that validates that the direct caller (predecessor account ID) matches to the contract's account (current account ID). The `({ privateFunction: true })` decorator simplifies this by making it a single line decorator while improving readability.
+
+Use this annotation within the designated contract class with the [`NearBindgen({})` decorator](../contract-structure/near-bindgen.md) as follows:
+
+```js
+@call({ privateFunction: true })
+my_method({}) {
+    // ...
+}
+```
+
+Which is equivalent to:
+
+```js
+@call({})
+my_method({}) {
+    if near.currentAccountId() != near.predecessorAccountId() {
+        throw new Error("Method method is private");
+    }
+    // ...
+}
+```
+
+Now with this annotation, only the account of the contract itself can call this method, either directly or through a promise.
+
+## Writing Internal Methods
+
+Not all functions need to be exposed publicly. It may be beneficial to write private methods for helper or utility functions, for instance. There are three approaches to write internal methods:
+
+1. Declare the method without using the `call` or `view` decorators.
+
+```js
+helperMethod(a, b) {
+  // ...
+}
+```
+
+2. Using an internal helper function in the module scope.
+
+```javascript
+// Function that can be called in another Rust file
+function getFirstName(account) {
+  // ...
+}
+```
+
+3. Importing a helper function or class from another module.
+
+Another way of not exporting methods is by having a separate class, that is not marked with `NearBindgen({})`.
+
+```js
+import { getFirstName } from "./helpers.js";
+
+@NearBindgen({})
+export class Contract {
+  // ...
+}
+
+class Helpers {
+  // ...
+}
+```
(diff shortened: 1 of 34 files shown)


idea404:
@gagdiez @bucanero @maxhr 

Please go ahead and lmk what you think. There are some sections that are left as TODOs for a next iteration, which I can make issues for once this gets merged:
- Custom build commands using the JS SDK CLI
- Add `UnorderedSet` and `LookupSet` explanation in `collections.md`
- Add Nesting section for collections
- Add handy workspaces snippets in testing section 
- Add Best Practices section
- Add reducing wasm file size section

docs/sdk/js/contract-structure/collections.md
```diff
@@ -0,0 +1,158 @@
+---
+sidebar_position: 2
+title: "Collections"
+---
+
+# Collections
+
+When deciding on data structures to use for the data of the application, it is important to minimize the amount of data read and written to storage but also the amount of data serialized and deserialized to minimize the cost of transactions. It is important to understand the tradeoffs of data structures in your smart contract because it can become a bottleneck as the application scales and migrating the state to the new data structures will come at a cost.
```
bucanero, line 8:
```suggestion
When deciding on data structures to use for the application's data, it is important to minimize the amount of data read and written to storage and the amount of data serialized and deserialized to minimize the cost of transactions. It is important to understand the tradeoffs of data structures in your smart contract because it can become a bottleneck as the application scales, and migrating the state to the new data structures will come at a cost.
```

docs/sdk/js/contract-interface/public-methods.md
```diff
@@ -0,0 +1,30 @@
+---
+sidebar_position: 1
+title: "Public Methods"
+---
+
+# Public Method Types
+
+Methods can be called externally by using the `view({})` or `call({})` decorators within the coontract class which will expose the method in the compiled WASM bytecode to be called by any other NEAR Account. Whenever a method is declared in your contract class without these decorators and is called by another NEAR Account, a `MethodNotFound` error will be thrown.
+
+If you need a contract to call itself, you can mark the function with these decorators but add the [`({  privateFunction: true })` annotation in the decorator parameters](private-methods.md) so that it will panic if called from anything but the contract itself.
```
bucanero, line 10:
```suggestion
:::tip

If you need a contract to call itself, you can mark the function with these decorators but add the [`({  privateFunction: true })` annotation in the decorator parameters](private-methods.md) so that it will panic if called from anything but the contract itself.

:::
```

docs/sdk/js/contract-interface/private-methods.md
```diff
@@ -0,0 +1,71 @@
...
+```js
+helperMethod(a, b) {
+  // ...
+}
+```
+
+2. Using an internal helper function in the module scope.
+
+```javascript
+// Function that can be called in another Rust file
```
bucanero, line 50:
```suggestion
// Function that can be called in another JS file
```

docs/sdk/js/cross-contract/callbacks.md
```diff
@@ -0,0 +1,86 @@
...
+## Calculator Example
+
+A callback method can be declared in your contract class as a regular method decorated with the `call({})` decorator. Be sure to pass in the `privateFunction: true` option to the decorator. This will ensure that the method is only callable by the contract itself.
+
+For example, let's assume the calculator is deployed on `calc.near`, we can use the following:
+
+```js
+@NearBindgen({})
+export class CalculatorCallerContract {
+  @call({})
```
bucanero, line 19: In the paragraph above, you mention `Be sure to pass in the 'privateFunction: true' option`, but I don't see the option in these code examples? 🤔

docs/sdk/js/cross-contract/callbacks.md
```diff
@@ -0,0 +1,86 @@
...
+    };
+}
+```
+
+After creating the class, we'll show a simple flow that will make a cross-contract call to the allowlist smart contract, asking if the account `idea404.testnet` is allowlisted. 
+
+```js
+@NearBindgen({})
+export class Contract {
+    @call({})
```
bucanero, line 57: same question about `privateFunction` ?

bucanero (approved):

bucanero:
One general question (for the future), do we want to keep both "Rust SDK" and "JavaScript SDK" navigation structures? or maybe we want at some point to unify both, and keep the same article content with a Rust tab (with Rust code snippets), and a JS tab (with Javascript code), as we have done for the Docs 3.0 update?

docs/sdk/js/cross-contract/callbacks.md
idea404: The callback method is `sum` belonging to the `calculatorAccountId` account. That method should have the decorator param. But let me add that here because I understand this is confusing.

docs/sdk/js/cross-contract/callbacks.md
idea404: In this case the example is below. I'll highlight in the text that setting the function to private is what you want to do for the callback method specifically.

idea404:
> One general question (for the future), do we want to keep both "Rust SDK" and "JavaScript SDK" navigation structures? or maybe we want at some point to unify both, and keep the same article content with a Rust tab (with Rust code snippets), and a JS tab (with Javascript code), as we have done for the Docs 3.0 update?

Hm I think they're quite different to one another, perhaps too different to merit fully merging them and going down the route you suggest. This PR makes them seem very similar but it also used the Rust template and took away the things that were too different and basically "ported" the overlaps. An example of something Rust SDK allows the dev to do which JS SDK does not is optimizing/reducing `.wasm` file size for example with flag customization for compilation.

PR #1314, opened 2022-12-06
move from Kurtosis modules to Kurtosis packages

h4ck3rk3y:
Kurtosis now creates Starlark packages that modify enclaves, this is the one for NEAR https://github.com/kurtosis-tech/near-package

This replaces https://github.com/kurtosis-tech/near-kurtosis-module


h4ck3rk3y:
@bucanero At Kurtosis we've migrated towards packages over the older modules, can we get these docs in please?

bucanero (approved):

PR #1323, opened 2023-01-06
Adding Filebase to list of IPFS Pinning Services

jamongeon:
Adding Filebase to list of IPFS Pinning Services


bucanero:
Since we merged #1358 , can we close this PR @jamongeon ?

PR #1330, opened 2023-01-13
Include Tip for Saving on Storage

idea404:
This PR includes a tip on how to save on storage costs when indexing by NEAR Account ID.

### DIFF
--- a/docs/2.develop/contracts/storage.md
+++ b/docs/2.develop/contracts/storage.md
@@ -217,6 +217,10 @@ Your contract needs to lock a portion of their balance proportional to the amoun
 
 Currently, it cost approximately **1 Ⓝ** to store **100kb** of data.
 
+:::info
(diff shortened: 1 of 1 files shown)


docs/2.develop/contracts/storage.md
```diff
@@ -217,6 +217,10 @@ Your contract needs to lock a portion of their balance proportional to the amoun
 
 Currently, it cost approximately **1 Ⓝ** to store **100kb** of data.
 
+:::info
+You can save on smart contract storage if using NEAR Account IDs by compressing them using base32 encoding. Since they consist of [a-z.-_] characters with a maximum length of 64 characters, they can be encoded using 5 bits per character, with terminal \0. Going to a size of 65 * 5 = 325 bits from the original (64 + 4) * 8 = 544 bits. This is a 40% reduction in storage cost.
```
bucanero, line 221: I'd remove the "compression" reference as base32/base64 are encoding algorithms and not compression algos. (Regular binary data will require more storage space if encoded with base64)

```suggestion
You can save on smart contract storage if using NEAR Account IDs by encoding them using base32. Since they consist of `[a-z.-_]` characters with a maximum length of 64 characters, they can be encoded using 5 bits per character, with terminal `\0`. Going to a size of 65 * 5 = 325 bits from the original (64 + 4) * 8 = 544 bits. This is a 40% reduction in storage costs.
```

bucanero (approved):

PR #1362, opened 2023-02-27
update RAM requirements for chunk-only producers

marcelo-gonzalez:
This updates the RAM requirements that should have been bumped after 1.28.1, to be consistent with https://github.com/near/node-docs/blob/main/docs/rpc/hardware-rpc.md


marcelo-gonzalez:
example of a similar change: https://github.com/near/node-docs/pull/53

bucanero (approved):

PR #1378, opened 2023-04-04
Added a section to describe the interaction with other macros.

micky-gee:
This addition describes some unintuitive behaviour that results from the near_bindgen macro expanding the extern function, that leads to all attribute-like macros that may decorate that function being executed twice.  

For an example, see: https://stackoverflow.com/questions/75887893/near-sdk-near-bindgen-colliding-with-other-macros


gagdiez:
Hi @micky-gee , thanks for the contribution. Maybe it would be better to add that section as an `details`, instead of a whole sub-section. 

Check: https://docusaurus.io/docs/next/markdown-features#details

micky-gee:
Thank you for your review, appreciated.

I'm glad that you accept the content and we've just got to find a way to format it appropriately.

Looking through the doc, I can't find other places where a "details" as you link it, is used.  I also note that the specific implementation isn't markdown, but rather some additional HTML.  Lets drop this in where its obvious to users who may be facing the same macro challenges I was; rather than hiding it behind some non-standard javascript widget?

gagdiez:
@micky-gee since the data is relevant to some users, and not all of them (only you reported this in 2 years), it would be better to have it optionally as a details component, which is optional to read, instead of a mandatory section for everyone.

Our website runs on Docusaurus, which is a mixture of markdown and other features. You can see examples of the details section here https://github.com/search?q=repo%3Anear%2Fdocs+%3Cdetails%3E&type=code

So I assume your code should look something like this:

```html
<details>
  <summary> Interaction with other macros </summary>

When `near_bindgen` is built for the wasm32 target, it generates the external NEAR contract bindings.  To achieve this it is actually generating another function with the signature `pub extern "C" fn function_name()` that first deserializes the contract struct from NEAR storage and then calls the `contract.function_name(parameter1, parameter2, ...)`.  If you have annotated your function with any attribute-like macros, these are then executed _twice_.  Specifically if the attribute like macro makes any modification to the function signature, or inserts any code that depends on the contract struct (in the form of `&self`, `&mut self`, or `self`) this will fail in the second invocation, because the externally exposed function does not have any concept of this struct.  It is possible to detect this by checking which build target you are building for and limit the execution of the macro to operate only on the first pass.

</details>
```

gagdiez:
@micky-gee on a further evaluation of your [stackoverflow question](https://stackoverflow.com/questions/75887893/near-sdk-near-bindgen-colliding-with-other-macros), the error is arises from using the rust sdk in a non-documented way.

This is, it does not follow any of our documentation or guidelines.

Our of curiosity, what were you trying to build?

docs/sdk/rust/contract-structure/near-bindgen.md
```diff
@@ -94,3 +94,7 @@ pub fn my_method(&mut self) {
 ...
 }
 ```
+
+## Interaction with other macros
```
gagdiez, line 98: This could be better as a detail. Thought I don't understand in which context a smart-contract developer would find this error.

gagdiez (changes requested):
This does not seam to be targeted to a broad smart contract developer audience.

micky-gee:
In many contract calls folks spend a fair amount of effort validating a set of conditions, for example:
1. That the caller (`env::signer_account_id`) has some property (for example if you're modifying some struct content, that the instance was created by the caller);
2. That some information already exists (i.e. the call is in order after a create and before a delete).

So I created some attribute macros to decorate the function call, and by means of an example:
```
    #[handle_result]
    #[state_safe(caller = "member")]
    pub fn transfer_to_participant(
```

or 
```
    #[handle_result]
    #[state_safe(caller = "owner_only")]
    pub fn terminate(
```

This allows for a clean separation of the validation rules from the contract logic.

Would love to hear your thoughts on this.

gagdiez:
@micky-gee Indeed, a non-standard way to use it. Let's add it as a `detail` then, since I do not expect most users to run into this problem.

micky-gee:
Okey dokey; thanks—will modify.

gagdiez (approved):

PR #1380, opened 2023-04-07
Discovery design components

bucanero:


calebjacob (changes requested):
Hey @bucanero, I think this is a great place to start and we can expand on it down the road. Just a couple minor observations:

- Maybe break out the context section into its own tutorial and let's add documentation about `context.accountId` for the current signed in account too.
- The `Built-in Components` should also probably reference Radix.
- At the bottom of the design tutorial where you're listing the components, maybe mention some details about `DIG.Theme` (it already wraps all of discovery so they don't need to render it themselves, but it means they can use any of the CSS variables defined inside).

bucanero:
hi @calebjacob 👋 

I've added some changes based on your feedback:
- moved "VM context" to the "Components" page, including `accountId` https://near-docs-pr-1380.onrender.com/discovery/home
- added a reference to Radix UI under the "Built-in Components" page: https://near-docs-pr-1380.onrender.com/discovery/components/home#non-exhaustive-list-of-built-in-components
- added some notes regarding `DIG.Theme`

Let's sync on Slack once you had a chance to check these changes. 👍

calebjacob (approved):
This looks great, thanks @bucanero!

PR #1383, opened 2023-04-13
feat(i18n): enable Korean language

think-in-universe:
This is to enable Korean language on docs as the NEAR Korea Hub has almost finished translating all the docs on https://docs.near.org.

A few errors about the Korean translation have been fixed: https://hackmd.io/@L9ZnGph8Tbaio2GVX_z-EA/B1-95HBfn

Let's do more testing before merge this into production.

### DIFF
--- a/.github/workflows/build-check.yml
+++ b/.github/workflows/build-check.yml
@@ -22,6 +22,6 @@ jobs:
         run: |
           cd website && yarn
           yarn crowdin:build
-          yarn build --locale zh-CN && yarn build --locale vi && yarn build --locale en
+          yarn build --locale zh-CN && yarn build --locale vi && yarn build --locale ko && yarn build --locale en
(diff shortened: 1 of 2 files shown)


bucanero:
@think-in-universe , we are using Crowdin for translations, and we can't push this content directly to the main repository.

This would break the whole process and how we handle updates and translations. When we update the main docs, we won't have any way to track the differences and outdated content on the translated pages.

They should commit all the content through Crowdin; that way the translations can be automatically added to our docs.

bucanero:
> I'm also looking at the build failure, which might come from some broken files on Crowdin. Will let you know for any update.

From our review on Slack, the only issue is related to `environment.md` having some broken `<Tabs>` content.
We'll update Crowdin's translated file and try to build again.

PR #1391, opened 2023-04-24
Fix links when using localized docs

bucanero:


website/components/featurelist.js
```diff
@@ -20,9 +21,13 @@ export function Column({ title, children, size = 4 }) {
 }
 
 export function Feature({ image, title, subtitle, url }) {
+  const { siteConfig, i18n } = useDocusaurusContext();
   let link;
   if (url.startsWith('http')) {
     link = <svg width="0.8rem" height="0.8rem" aria-hidden="true" viewBox="0 0 24 24" class="iconExternalLink_node_modules-@docusaurus-theme-classic-lib-theme-Icon-ExternalLink-styles-module"><path fill="currentColor" d="M21 13v10h-21v-19h12v2h-10v15h17v-8h2zm3-12h-10.988l4.035 4-6.977 7.07 2.828 2.828 6.977-7.07 4.125 4.172v-11z"></path></svg>;
+  } else {
+    if (i18n.currentLocale != i18n.defaultLocale)
```
think-in-universe, line 29: I suggest explicitly adding braces to avoid mistakes, in case we need to add more lines to the code block in future. 

```js
    if (i18n.currentLocale != i18n.defaultLocale) {
      url = "/" + i18n.currentLocale + url;
    }
```

or simply move the block into `else if`

```js
  if (url.startsWith('http')) {
    link = <svg width="0.8rem" height="0.8rem" aria-hidden="true" viewBox="0 0 24 24" class="iconExternalLink_node_modules-@docusaurus-theme-classic-lib-theme-Icon-ExternalLink-styles-module"><path fill="currentColor" d="M21 13v10h-21v-19h12v2h-10v15h17v-8h2zm3-12h-10.988l4.035 4-6.977 7.07 2.828 2.828 6.977-7.07 4.125 4.172v-11z"></path></svg>;
  } else if (i18n.currentLocale != i18n.defaultLocale) {
    url = "/" + i18n.currentLocale + url;
  }
```

think-in-universe (changes requested):
Thanks Damian. 

I have tested the fix locally and it works perfectly. The i18n paths have been added. 😄 

One suggestion about the code as I commented.

website/components/featurelist.js
bucanero: fixed 👍

PR #1393, opened 2023-04-25
update near.org links

bucanero:


mpeterdev (changes requested):
I just noticed two more necessary changes. All instances of `adminalpha.near` should now just be `near` and `calebjacob.near/widget/ComponentDetailsPage` should be `near/widget/ComponentDetailsPage`

mpeterdev (approved):
🎉

PR #1397, opened 2023-05-04
Update BOS index.md to match Docs site

bucanero:


gagdiez:
Damn... I completely missed this PR ::palm-face::

I opened https://github.com/near/docs/pull/1400, would suggest to continue in #1400 since it has more changes added

PR #1403, opened 2023-05-25
docs: use bos loader on testnet

think-in-universe:
As suggested by discussion in NEAR Social dev telegram (https://t.me/NearSocialDev/3327), add instructions about how to use BOS loader on testnet.

### DIFF
--- a/docs/bos/dev/bos-loader.md
+++ b/docs/bos/dev/bos-loader.md
@@ -35,6 +35,12 @@ In this article you'll learn how to develop, test, and deploy BOS components usi
 
 8. Open `https://near.org/<youraccount.near>/widget/<component name>` (case sensitive)
 
+   :::info
+   If you're testing on testnet, use your testnet account, and open https://test.near.org instead.
+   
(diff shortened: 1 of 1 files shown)


think-in-universe:
Just curious why CI build-check failed with `- Required option 'api_token' is missing`, I'm pretty sure it should have been configured.

bucanero:
> Just curious why CI build-check failed with `Required option 'api_token' is missing`. I'm pretty sure it should have been configured.

hi @think-in-universe , I have seen this, and I think it's related to some kind of security permissions and token visibility. For example, if I open a PR and I'm a member of the NEAR github group, then the CI build works fine and the API token is there.
But if someone from outside the NEAR github org opens a PR, then the token is not visible and the CI fails.

docs/bos/dev/bos-loader.md
```diff
@@ -35,6 +35,12 @@ In this article you'll learn how to develop, test, and deploy BOS components usi
 
 8. Open `https://near.org/<youraccount.near>/widget/<component name>` (case sensitive)
 
+   :::info
+   If you're testing on testnet, use your testnet account, and open https://test.near.org instead.
```
bucanero, line 39:
```suggestion
   If you're testing on `testnet`, use your testnet account and open https://test.near.org instead.
```

bucanero (approved):

PR #1411, opened 2023-06-19
Update access-keys.md

mlibre:
As I know it is also possible to create long named top-level accounts using `Full access keys`. and it is not limited to `sub-accounts`

### DIFF
--- a/docs/1.concepts/basics/accounts/access-keys.md
+++ b/docs/1.concepts/basics/accounts/access-keys.md
@@ -45,7 +45,7 @@ As the name suggests, `FullAccess` keys have full control of an account, similar
 
 Particularly, Full Access keys can be used to sign transactions doing [any action](https://nomicon.io/RuntimeSpec/Actions) in your account's behalf:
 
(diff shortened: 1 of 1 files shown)


BenKurrek (approved):

docs/1.concepts/basics/accounts/access-keys.md
```diff
@@ -45,7 +45,7 @@ As the name suggests, `FullAccess` keys have full control of an account, similar
 
 Particularly, Full Access keys can be used to sign transactions doing [any action](https://nomicon.io/RuntimeSpec/Actions) in your account's behalf:
 
-1. Create immediate [sub-accounts](account-id.md#rules-for-creating-named-accounts).
+1. Create immediate [sub-accounts](account-id.md#rules-for-creating-named-accounts) and [top-level](account-id.md#named-accounts) accounts if the length of the account ID is greater than 31 characters.
```
bucanero, line 48:
```suggestion
1. Create immediate [sub-accounts](account-id.md#rules-for-creating-named-accounts) and [top-level accounts](account-id.md#named-accounts) if the account ID's length is 32 characters.
```

PR #1412, opened 2023-06-21
Ordering Data section articles

mlibre:
The **Data on the Blockchain** section is a good place to start because it provides a general overview of Data on blockchains. This will give the reader a good foundation before they move on to the more specific topics of **NEAR Data Flow** and **Token Transfer Flow**.

The **NEAR Data Flow** section and the **Token Transfer Flow** section are related, so it makes sense to put them together. The **Token Transfer Flow** section even references the **NEAR Data Flow** section as it is previous article, so it is helpful to have those two sections in close proximity.

image

### DIFF
--- a/docs/1.concepts/data-flow/data-storage.md
+++ b/docs/1.concepts/data-flow/data-storage.md
@@ -1,7 +1,6 @@
 ---
 id: data-storage
 sidebar_label: "Data on the Blockchain"
-sidebar_position: 1
 ---
 
 # Data on the Blockchain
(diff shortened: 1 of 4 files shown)


bucanero:
@mlibre , if you want to change the sort order, you should edit the `sidebars.json` file

mlibre:
> @mlibre , if you want to change the sort order, you should edit the `website/sidebars.json` file

Oh. did not know. fixed!

bucanero (approved):

PR #1453, opened 2023-07-30
Update 6-royalty.md

dhsimpson:
This is an example related to royalty tokens, but there are some near cli calls and explanations related to tokens that are incorrectly labeled as 'approval,' so I will correct them to 'royalty.


bucanero (approved):

bucanero:
Thanks for your contribution @dhsimpson 👍

PR #1475, opened 2023-09-02
Update lock.md

ghimire007:
updated command to use near-cli-rs instead of JS version


gagdiez:
Same comment as #1476

PR #1476, opened 2023-09-02
Update deploy.md

ghimire007:
use near-cli-rs instead of JS version

### DIFF
--- a/docs/2.develop/deploy.md
+++ b/docs/2.develop/deploy.md
@@ -26,20 +26,19 @@ Thanks to the `NEAR CLI` deploying a contract is as simple as:
 #### Create an Account and Deploy
 ```bash
 # Automatically deploy the wasm in a new account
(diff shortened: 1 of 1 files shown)


gagdiez:
While I agree that we need to add more information on the updated `near-cli-rs`, we need to be careful on how we do it. There has been no announce that the classic `near cli` is being deprecated, and even if there was, we cannot just eradicate de old tool since lots of people rely on it.

It is better to have only one PR to add near CLI across all docs, and to do so using the same "tabs" system we have for the SDKs.

ghimire007:
This changes were made in response to raised issue [here](https://github.com/near/near-cli-rs/issues/235). It makes sense having tab based system like we have for js and rs sdk. what your take on this @frol ?

frol:
Sure, let's go with tabs. In the future we will add `cargo-near` where applicable and make them the default where it makes sense.

PR #1478, opened 2023-09-03
Added Tabs for  both  near-cli-rs  and  near-cli  commands 

ghimire007:
As discussed in  https://github.com/near/docs/pull/1476.
Docs are updated with tabs for both near-cli and near-cli-rs commands .
Commands mentioned in https://github.com/near/near-cli-rs/issues/236 are yet to be changed .


bucanero:
You should run and check locally if everything works.

In this case, you need to import the `Tabs` components or else it won't show up:
```js
import Tabs from '@theme/Tabs';
import TabItem from '@theme/TabItem';
```

bucanero (changes requested):
Fix Tabs imports

PR #1480, opened 2023-09-04
Added Tabs for both near-cli-rs and near-cli commands

ghimire007:
As discussed in https://github.com/near/docs/pull/1478 tabs imports are fixed and Commands mentioned in https://github.com/near/near-cli-rs/issues/236  are updated


gagdiez:
@ghimire007 could you rename all instances of `Near-CLI` to `near-cli`? after that. Also, make sure to remove all unnecessary new lines. After that I'll accept it and merge it. Thanks for the contribution!

gagdiez (approved):

PR #1500, opened 2023-09-21
add QueryApi context object information

bucanero:


docs/bos/queryapi/context.md
```diff
@@ -0,0 +1,265 @@
...
+    post_id: postId
+  }
+);
+```
+
+In this example, any rows where account_id and post_id match the supplied value are deleted. All deleted rows are returned. 
+
+### Auto Complete
+
+The above methods are generated when the schema is read. The tables in the schema are parsed and methods set under each table name. This makes using the object more intuitive and declarative. In addition to that, typescript types are generated which represent the table itself. These types are set as the parameter’s types. This provides autocomplete and strong typing in the IDE. These restrictions are not enforced on the runner side and are instead mainly as a suggestion to help guide the developer to use the methods in a way that is deemed correct by us. Here’s some screenshots of autocomplete and strong typing in action:
```
darunrs, line 263: Pavel gave some good advice that we should highlight that the autocomplete works while writing the schema and before publishing to the chain. In other words, you don't need to publish the indexer to get autocomplete. 

Some additions of my own: 
Right now, indexers by default had a context object included as a parameter in the top-level "async function getBlock". This blocks autocomplete as the local context object will shadow the global one, preventing access to it. Customers will need to manually remove the context paramter from their indexers and they will then see autocomplete. 

`async function getBlock(block: Block, context) {`

should become

`async function getBlock(block: Block) {`

I also added a button next to the format button. It has a "<>" symbol in it. That manually generates the types. It can be useful for debugging and iterative development while writing the schema.

darunrs: Also forgot to add, if the schema is invalid for generating types, then an error will pop up both on screen and in the console. Here's a picture of that. 
image

image

bucanero: Hi @darunrs , thanks for the additional notes and screenshots

I have updated the PR , adding all these notes. Please review it when you have a chance 👍

docs/bos/queryapi/context.md
```diff
@@ -0,0 +1,348 @@
...
+
+## Main Methods
+
+:::tip
+Most methods are asynchronous, hence why all examples have the `await` keyword in front of the function call. 
+:::
+
+### GraphQL
+
+When an indexer is published, the DDL supplied is used to spin up a Postgres database. This database is hooked up to Hasura, a data platform that supplies a GraphQL endpoint that can be used to interact with the Postgres database that was spun up.
```
darunrs, line 25: Maybe we can change the wording to "the SQL DDL written under the 'schema.sql' tab is used to spin up a postgres database. This database is integrated with Hasura to provide a graphQL endpoint. This endpoint can be used to interact with your database.".

docs/bos/queryapi/context.md
```diff
@@ -0,0 +1,348 @@
...
+:::caution Under development
+
+Many of the formatting and changes in this document are still in progress. Once these changes are merged, they will be available in Dev. Specifically, [auto-complete](#auto-complete) and the `context.db.TableName.methodName` format. 
+
+:::
+
+## Main Methods
+
+:::tip
+Most methods are asynchronous, hence why all examples have the `await` keyword in front of the function call.
```
darunrs, line 20: Actually all of them are. I double checked.

docs/bos/queryapi/context.md
```diff
@@ -0,0 +1,348 @@
...
+#### Example
+
+```js
+const h = block.header().height;
+await context.set("height", h);
+```
+
+### Log
+
+Code written by developers is executed on a GCP compute instance. Therefore, things like `console.log` are surfaced to the machine itself. So, to surface `console.log` statements not only to the machine but also back to the user under Indexer Status, we need to write these logs to the logging table, which is separate from the developer's schema. This actually happens in the runner. We map `console.log` statements in the developer's code to call this function instead. So, for the developer, they can simply use `console.log`, and we take care of the rest.
```
darunrs, line 97: The person reading this is going to be the developer in question. I wrote most of what I did with the audience being our team. So, some paragraph slike this one could probably be re-written to read nicely if I am a customer. 

On that note, there's no additional functionality provided to users through this method. Maybe we're better off not even mentioning it in the doc?

docs/bos/queryapi/context.md
```diff
@@ -0,0 +1,348 @@
...
+
+:::
+
+## DB
+
+The DB object is a sub-object under `context`. It is accessed through `context.db`. Previously, the GraphQL method was the only way to interact with the database. However, writing GraphQL queries and mutations is pretty complicated and overkill for simple interactions. So, simpler interactions are instead made available through the db sub-object. This object is built by reading the schema written by the developer, parsing its information, and generating methods for each table. See below for what methods are generated for each table. The format to access the below methods is as follows: `context.db.[tableName].[methodName]`. Concrete examples are also given below. 
+
+:::info Note
+
+One thing to note is that the process where the code is read is not fully featured. If `ALTER TABLE ALTER COLUMN` is used, for example, in the schema, it will fail to parse. Should this failure occur, the context object will still be generated but `db` methods will be unavailable. An error will appear on the developer’s page saying types were not generated.
```
darunrs, line 144: We can rewrite the second line to: 
If an `ALTER TABLE ALTER COLUMN` statement is used in the SQL schema, for example, it will fail to parse.

Also, like mentioned in my previous comment, lines like 
`An error will appear on the developer’s page saying types were not generated`
 should be rewritten to not talk as though the customer is not the one reading it. That line should be something like 
`An error will appear on the page saying types could not be generated. A more detailed error can be viewed in the browser console.`

docs/bos/queryapi/context.md
```diff
@@ -0,0 +1,348 @@
...
+
+should become
+
+```js
+async function getBlock(block: Block) {
+```
+
+#### Examples
+
+Some screenshots of autocomplete and strong typing in action:
```
darunrs, line 338: Maybe rephrase this to something more direct? Like: "Here are some screenshots which demonstrate autocomplete on methods, strong typing, and field names."

darunrs (commented):
Overall really nicely done! I love the addition of bubble text for things like tips. It feels very engaging. Some specific places I think could be reworded. In general, I wrote the stuff up with our team as the audience in mind. This document has a different audience in mind so some sentences read a bit clunky. Maybe a quick pass through to detect any weird language that talk about the developer/customer like they're not the one reading the doc would go a long way, I think.

darunrs:
Also, I see that there are two dead links. You can see them under the details of the failing check. They're not on pages we wrote, I think. But we probably want to at least reach out to whoever added it to let them know. Maybe we can update them ourselves if they send us the updated links?

pkudinov:
Resolves https://github.com/near/queryapi/issues/233

docs/bos/queryapi/context.md
```diff
@@ -0,0 +1,348 @@
+---
+id: context-object
+title: QueryAPI Context object
+sidebar_label: Context object
+---
+
+## Overview
+
+The `context` object is a global object made available to developers building indexers with [QueryAPI](intro.md). It provides helper methods for developers to interact with things outside their coding environment. Crucially, these involve the database and GraphQL endpoint spun up along with the indexer, as well as allowing API calls.
```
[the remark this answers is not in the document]
bucanero: fixed

docs/bos/queryapi/context.md
```diff
@@ -0,0 +1,348 @@
...
+sidebar_label: Context object
+---
+
+## Overview
+
+The `context` object is a global object made available to developers building indexers with [QueryAPI](intro.md). It provides helper methods for developers to interact with things outside their coding environment. Crucially, these involve the database and GraphQL endpoint spun up along with the indexer, as well as allowing API calls. 
+
+:::caution Under development
+
+Many of the formatting and changes in this document are still in progress. Once these changes are merged, they will be available in Dev. Specifically, [auto-complete](#auto-complete) and the `context.db.TableName.methodName` format.
```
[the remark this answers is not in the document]
bucanero: got it, I'm updating this "alert box"

docs/bos/queryapi/context.md
bucanero: fixed

docs/bos/queryapi/context.md
bucanero: fixed

docs/bos/queryapi/context.md
bucanero: Yes, I agree that we can remove the whole `Logs` section since it does not provide any helpful info for the developers reading it.

bucanero: - section removed -

docs/bos/queryapi/context.md
bucanero: fixed

docs/bos/queryapi/context.md
bucanero: fixed

docs/bos/queryapi/context.md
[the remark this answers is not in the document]
bucanero: I updated it as:

> The formatting and changes in this document are still in progress. These changes are not fully featured yet but will be by the time they hit production. Specifically, [auto-complete](#auto-complete) and the `context.db.TableName.methodName` format.

bucanero:
> Also, I see that there are two dead links. You can see them under the details of the failing check. They're not on pages we wrote, I think. But we probably want to at least reach out to whoever added it to let them know. Maybe we can update them ourselves if they send us the updated links?


Since the links are not from this PR, I think we can make another PR to review and fix broken links. I think this shouldn't be a blocker to release this useful queryApi article.

darunrs (approved):

PR #1526, opened 2023-10-06
Update context.md - add log

bucanero:


docs/bos/queryapi/context.md
```diff
@@ -92,6 +92,10 @@ const h = block.header().height;
 await context.set("height", h);
 ```
 
+### Log
+
+Your indexing code is executed on a GCP compute instance. Therefore, things like `console.log` are surfaced to the machine itself. To surface `console.log` statements not only to the machine but also back to you under _Indexer Status_, QueryAPI needs to write these logs to the logging table, which is separate from the developer's schema. This [happens in the runner](how-works.md), and QueryAPI maps `console.log` statements in the developer's code to call this function instead, so you can simply use `console.log`, and QueryAPI takes care of the rest.
```
darunrs, line 97: Can we update this to state it runs on a cloud compute instance? GCP is only used for V2. In V1, the code is run in AWS Lambda.

darunrs (changes requested):
In the future, maybe we can reword this doc to either fully use you/your or developer. We're mixing the two right now, which reads a little awkwardly. Doesn't have to happen in this PR. 

Once the below is addressed, I can approve.

docs/bos/queryapi/context.md
bucanero, line 97:
```suggestion
Your indexing code is executed on a cloud compute instance. Therefore, things like `console.log` are surfaced to the machine itself. To surface `console.log` statements not only to the machine but also back to you under _Indexer Status_, QueryAPI needs to write these logs to the logging table, which is separate from the developer's schema. This [happens in the runner](how-works.md), and QueryAPI maps `console.log` statements in the developer's code to call this function instead, so you can simply use `console.log`, and QueryAPI takes care of the rest.
```

bucanero: fixed 👍

bucanero:
> In the future, maybe we can reword this doc to either fully use you/your or developer. We're mixing the two right now, which reads a little awkwardly. Doesn't have to happen in this PR.
> 
> Once the below is addressed, I can approve.

Yes, you're right. Let me do a review and adjust the content accordingly.

darunrs (approved):

PR #1544, opened 2023-10-17
feat: BOS for Ethereum developers doc

zavodil:


bucanero:
@zavodil , since this is a new article, I suggest to edit `website/sidebar.json` too and add the article there, so it shows up in the left navigation

bucanero:
> @bucanero will do. Do you already have a category like "Building on BOS"? If no, what is the best place to put it?

yes, we have a section for BOS here: https://docs.near.org/develop/integrate/welcome

I think the best option would be to replace the following "Basic ether.js" article:
https://docs.near.org/bos/tutorial/ethers-js

since that old article feels like a place holder, your new article would be much better for devs to learn how to use BOS+Etherjs

website/sidebars.json
```diff
@@ -618,7 +618,7 @@
         "bos/tutorial/using-iframes",
         "bos/tutorial/bos-gateway",
         "bos/tutorial/push-notifications",
-        "bos/tutorial/ethers-js",
+        "bos/tutorial/bos-ethersjs",
```
bucanero, line 621:
```suggestion
        "bos/tutorial/ethers-js",
```

docs/bos/tutorial/bos-ethersjs.md
```diff
@@ -0,0 +1,145 @@
+---
+id: bos-ethersjs
```
bucanero, line 2:
```suggestion
id: ethers-js
```

bucanero:
@zavodil , could you remove the old file `hello-ethersJS.md`? thank you! 🙏

zavodil:
@bucanero done

PR #1546, opened 2023-10-17
Add a tutorial for websockets based on the widget activity

bucanero:


docs/bos/tutorial/queryapi-ws.md
```diff
@@ -0,0 +1,344 @@
...
+  }
+```
+
+## Using WebSockets
+
+Once you have a QueryAPI indexer running, you can use WebSockets to get the data in your B.O.S. component. You only need to create a `WebSocket` object pointing to the QueryAPI's GraphQL endpoint.
+
+### Setup
+
+Here's a code snippet from the BOS component that subscribes and processes any activity from the [Widget Activity indexer](#queryapi-indexer):
```
roshaans, line 152: I tried to copy and paste the code into the editor to find out that it does not display anything. 

Can we create a note saying that this is just a snippet. and if you would like to play around with the widget, you can see it at this link and fork it?

docs/bos/tutorial/queryapi-ws.md
```diff
@@ -0,0 +1,344 @@
...
+    );
+  const newActivities = [
+    ...incoming_widgetActivities.filter((activity) => {
+      return (
+        state.widgetActivities.length == 0 ||
+        activity.block_timestamp > state.widgetActivities[0].block_timestamp
+      );
+    }),
+  ];
+  if (newActivities.length > 0 && state.widgetActivities.length > 0) {
```
roshaans, line 281: You may just want to get rid of anything to do with sounds as it does not work haha

roshaans (approved):
small changes left. 

As a User of the docs, I want to take snippets of codes and naturally render them in the BOS, but when I did that it did not work as I found out it was a very minimal example.

docs/bos/tutorial/queryapi-ws.md
bucanero: removed 👍

docs/bos/tutorial/queryapi-ws.md
bucanero: Note added 👍

PR #1552, opened 2023-10-24
QueryApi best practices

bucanero:


ailisp (commented):
Nice writing! I think you've covered all my points and express them in a better, easier to understand manner. Although I'm not sure it's best practice, given there's no other suggested workflows, it can be best practice so far and until someone enhances the practice :D

ailisp (approved):

PR #1573, opened 2023-11-10
fix: resolve broken links

thisisjoshford:
Resolve broken links in docs.

- Removed mirrored TypeDocs for NAJ and forwarded users to the live TypeDocs site 
- Fixed bos-gateway repo link
- Removed inaccurate "tip" in primitives to non-existent content


bucanero (approved):

bucanero:
the last Typedocs to remove would be `near-sdk-js`, once that one is gone it should simplify a lot our docusaurus

PR #1587, opened 2023-11-15
Update near-bindgen.md

flmel:
Fix documentation for contract initialization with the js sdk `init` method does not exist and its actually `intialize`


thisisjoshford (approved):
nice catch! Thanks for the fix :)

PR #1598, opened 2023-11-23
chore: typos, markdown linting & refer to cli-rs when suggesting near cli tool

shariffdev:


gagdiez:
the big scope of this PR makes it very difficult to merge it, I would suggest to divide it into smaller PRs

PR #1618, opened 2023-12-12
fix: resolve fetching remote content via plugin

thisisjoshford:
**Issue:**

[This commit disables remote syncing](https://github.com/near/docs/commit/cb7e34abead2ad59f551fbfc086323bfd1a78fe0#diff-bc4913d026ae51bef2bc8b7a1c040fdf52b999d99e6fa40d87b379f98a60e0eaR105)  and requires a CLI command to check for updates. 
_(Changing `noRuntimeDownloads` to `true`)_

@gagdiez I believe you created the `check-logs` script to implement the CLI check but unfortunately it does not work as expected.  Running `getChangelogs.js` only updates an array of strings (markdown file names) that tells the `remote-content` plugin which files to look for. The CLI command requires running: [`docusaurus download-remote`](https://github.com/rdilweb/docusaurus-plugin-remote-content?tab=readme-ov-file#cli-sync) for each section

**This PR:**

- fixes `check-logs` script in `package.json` (now called `remote-sync`)
- updates reports w/ updated content (email subscription)

**TODO:** 

- Create a CI pipeline that runs `remote-sync` on a regular cadence 
- Consider importing `near-releases` as a submodule for this repo

### DIFF
--- a/website/blog/2023-01-31-changelog.md
+++ b/website/blog/2023-01-31-changelog.md
@@ -2,6 +2,8 @@
 
 Developer changelog for [essential repositories](/develop/github-overview) when building on NEAR Protocol. 🏗️
 
+👉 [Get monthly emails of this report](https://docs.google.com/forms/d/1JfFUbTq3ELUlScJT1UI9PQPuQsv0W2jcTa7P94KrS5U/edit) 👈
+
 ## Releases  🎉
 
 | repo | release | release_date |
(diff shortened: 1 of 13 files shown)


bucanero:
Maybe off-topic, but I think we should change the URL path from `/blog` to `/changelog` , as it feels weird that the URL and the content don't match.

thisisjoshford:
> Maybe off-topic, but I think we should change the URL path from `/blog` to `/changelog` , as it feels weird that the URL and the content don't match.

I actually tried renaming this before the original implementation but the blog feature requires this directory namespace. This was a workaround for the hackathon but certainly can look into a better implementation of this.

gagdiez (approved):

PR #1626, opened 2023-12-13
fix: css navbar & CI

gagdiez:
Making the navbar collapse before since now it has too many elements


bucanero (approved):

bucanero:
@gagdiez I see we have both `bootstrap.css` and `custom.css` files again. Is that right? Just flagging in case we "revive" the previous bug

gagdiez:
Indeed, I am using this PR to test the `bootstrap` issue. Will wait for the render and then remove it if necessary.

thisisjoshford (approved):

thisisjoshford:
@gagdiez As you mentioned the main nav becoming crowded, should we consider nesting `Primitives` under `Build`?

PR #1633, opened 2023-12-26
Update quick-reference.md

igorperic17:
Removing walledUrl from Network object init as it doesn't exist.
image
image


gagdiez:
Hi @igorperic17, thanks for the contribution. I was seeing that the `near-api-js` typedocs show that a `walletURL?` parameter exists: https://near.github.io/near-api-js/interfaces/near_api_js.connect.ConnectConfig.html

I wonder why you are seeing an error message.

gagdiez:
This has been stalled for 3 weeks, please open again if relevant

PR #1641, opened 2024-01-02
Remove mentions to BOS

gagdiez:
The "Blockchain Operating System" and its derived terminology has caused more harm than good. This PR starts a new effort to remove all mentions to BOS and shift our main narrative to "NEAR is the Gateway to an Open Web".


bucanero:
awesome change 👍 
side note, we also need to search/replace mentions to `B.O.S.` (I've seen some of those on our docs too)

bucanero (approved):

PR #1675, opened 2024-01-20
fix composing component widget

Devendra116:
The widget loads component from testnet, thus require to update the widget code


gagdiez:
This page should not be loaded anymore, since [`/bos/componets`](https://docs.near.org/bos/components) now redirects to https://docs.near.org/bos/api/builtin-components

bucanero:
> This page should not be loaded anymore, since [`/bos/componets`](https://docs.near.org/bos/components) now redirects to https://docs.near.org/bos/api/builtin-components

@gagdiez the redirect is not working, I have tested and it's opening the old page, with no sidebar on the left.

I think we probably need to remove this markdown file to get the redirect working.

gagdiez:
fixed in #1697

PR #1702, opened 2024-02-01
fix: Removed unnecessary css

gagdiez:
Removed more than 800 lines from our CSS that I believe were contributing nothing to the docs. Locally, it works perfectly, lets wait until the preview is ready to see if effectively everthting is fine.


bucanero (approved):
Preview link looks ok on my browser (firefox)

PR #1704, opened 2024-02-01
Add chain signatures article ("what it is")

bucanero:


docs/1.concepts/account-abstraction/chainsig.md
```diff
@@ -0,0 +1,114 @@
...
+Chain Signatures are a method to allow a smart contract to call an async method. For example:
+
+```rust
+async fn sign(
+    payload: Bytes,
+    key_derivation: Bytes,
+    key_type: ECDSA) -> Signature
+```
+
+The signature is calculated on the data payload using a user-provided key. The signature output is always the same, as long as the caller and payload remains unchanged.
```
DavidM-D, line 19: Interestingly the signatures aren't actually the same, they're different but both valid.
```suggestion
The signature is calculated on the data payload using a user-provided key. The signature's public key is always the same, as long as the caller and the key path remain the same.
```

docs/1.concepts/account-abstraction/chainsig.md
```diff
@@ -0,0 +1,114 @@
...
+
+:::info
+The signature is an ECDSA signature on the Secp256k1 curve.
+:::
+
+## Use cases
```
DavidM-D, line 25: One of my concerns here is that while these use cases are still valid, they're fairly convuluted and worse they're pretty remote from how you should actually build these smart contracts.

docs/1.concepts/account-abstraction/chainsig.md
```diff
@@ -0,0 +1,114 @@
...
+
+This is useful for paid content, whatsapp messages and any private data on chain.
+
+Apart from the obvious this also enables:
+- **Secret-bid auctions with MEV protection** Users can submit bids that are encrypted at the end of the auction, the smart contract can decrypt all bids with a single evaluation.
+- **Dead man’s switch**. Journalists or whistleblowers can ensure that compromising information in their possession is automatically published if they were to become incapacitated
+- **One-time programs** Programs that can be executed only once on a single input, and that don’t leak anything about the program other than the result of the computation
+
+
+### On Chain Domain Control[^2]
```
DavidM-D, line 106: Probably best to drop this, I still haven't had the time to prove that this is actually feasible.

DavidM-D (changes requested):
I feel a bit bad, I provided you with some resources that I wrote, you copied them fairly verbatim, then I complain in the reviews that what you wrote is factually inaccurate. Sorry since the original draft of this features some things have changed/become more clear.

docs/1.concepts/account-abstraction/chainsig.md
bucanero: Sure, these two were commented out. I'll remove them from the draft.

docs/1.concepts/account-abstraction/chainsig.md
bucanero: We can change /replace these use cases with others that fit better. We can share some ideas during our next call.

bucanero: Hi @DavidM-D , following up on this draft, do you want to replace any of the current use-cases, or want to add some new ones?

Also, let me know if you want to add more information to this article. (please link reference documents so I don't end up using outdated info)

docs/1.concepts/account-abstraction/chainsig.md
```diff
@@ -0,0 +1,124 @@
+---
+id: chain-signatures
+title: Chain Signatures
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+
+Chain Signatures allows you to call the `sign()` method on `multichain.near`:
```
thisisjoshford, line 10: I want to add more context to what chain signatures are here. Only having that they allow you to call a method on a contract isn't super helpful here. I've started a doc to help with this here @bucanero: https://docs.google.com/document/d/1Rxwg24SX6zT1VKdUkieeCHLlvUcJwUPHuC21XG1gUrQ

bucanero: Thanks for sharing, I've updated the draft and added more context about Chain signatures. 👍

docs/1.concepts/account-abstraction/chainsig.md
DavidM-D: I think adding in key markets is a good idea, so the ability to change the ownership of Bitcoin/Ethereum accounts to trade assets across chains without doing on chain transactions

bucanero: Hi @DavidM-D , I have added a section about trading assets without transactions. 

Preview link:
https://docs-preview-pr-1704.onrender.com/concepts/chain-abstraction/signatures#trade-blockchain-assets-without-transactions

docs/1.concepts/abstraction/relayers.md
```diff
@@ -0,0 +1,35 @@
...
+The user expresses: "I want to do a specific action on chain" and signs this intent **off-chain**, but does not sends it to the network. Instead, they send the intent to a `Relayer`, which wraps the message into an actual transaction, attaches the necessary funds, and sends it to the network.
+
+<details>
+<summary> Technical Details </summary>
+
+Technically, the end user (client) creates a `SignedDelegateAction` that contains the data necessary to construct a `Transaction`, signs the `SignedDelegateAction` using their key, and send it to  the relayer service.
+
+When the request is received, the relayer uses its own key to sign a `Transaction` using the fields in the `SignedDelegateAction` as input to create a `SignedTransaction`.
+
+The `SignedTransaction` is then sent to the network via RPC call and the result is sent back to the client. The `Transaction` executes in such a way that the GAS fees are payed by the relayer, but all actions execute as if the user had sent the transaction.
```
bucanero, line 23:
```suggestion
The `SignedTransaction` is then sent to the network via RPC call, and the result is sent back to the client. The `Transaction` is executed in such a way that the relayer pays the GAS fees, but all actions are executed as if the user had sent the transaction.
```

docs/1.concepts/abstraction/chain-signatures.md
```diff
@@ -0,0 +1,112 @@
...
+_Diagram of a chain signature in NEAR_
+
+<hr class="subsection" />
+
+### 1. Create Payload
+
+The first step is to construct a payload (transaction, message, data, etc.) for the target blockchain platform.
+
+This can be performed in the following steps:
+<!-- TODO -->
```
mikedotexe, line 35: can't wait to dive in! looking solid!

gagdiez (approved):

gagdiez:
@DavidM-D this is block on your review request, please review again and merge if you find it informative enough.

docs/1.concepts/abstraction/chain-signatures.md
```diff
@@ -4,22 +4,23 @@ title: Chain Signatures
...
 ---
 
-import Tabs from '@theme/Tabs';
-import TabItem from '@theme/TabItem';
+Chain Signatures unlock the ability for a single account to transact across multiple blockchain protocols, giving ownership of cross-chain accounts, data, and assets to one NEAR account.
 
-Chain Signatures unlock the ability for a single account to transact across multiple blockchain protocols, giving ownership of many accounts, data, and assets to one NEAR account. 🤯
+This many-to-one ownership is made possible through a mixture of services across our tech stack:
 
-This many-to-one ownership is made possible through NEAR's [unique account model](../basics/accounts/introduction.md), which allows many keys to be associated with a single account. Chain Signatures generates and uses one of these keys to sign payloads compatible with other blockchain platforms but still linked to a native NEAR account.
```
thisisjoshford: @gagdiez, why remove this section? I feel it's essential to link the account abstraction piece here and how the ability for one account to possess many keys makes this whole approach possible. Eth devs are used to a 1 key 1 account and helping them understand a new mental model is crucial for getting this concept.

docs/1.concepts/abstraction/chain-signatures.md
```diff
@@ -4,22 +4,23 @@ title: Chain Signatures
...
 
 ## How It Works
 
-Chain Signatures in four basic steps:
+There are four steps involved on Chain Signatures:
 
-1. [Create Payload](#1-create-a-payload) - A payload is created to be signed and sent to a given blockchain platform.
-2. [Signature Request](#2-request-signature) - A smart contract call is made to sign the payload.
-3. [MPC Signing Service](#3-sign-with-mpc) - A service signs the payload linking it with the sender's NEAR account.
+1. [Create a Transaction](#1-create-a-payload) - The user creates the transaction they intent to submit to another blockchain
```
thisisjoshford, line 21: @gagdiez, I used payload here as users are not limited to this one action. Messages, data, identity verification, etc. are also valid use cases. I understand that you are most likely going for simplicity here but worry we might be not technically correct or introducing a perceived limitation to devs. 

@volovyks @DavidM-D what are your thoughts on using transactions over payload?

docs/1.concepts/abstraction/chain-signatures.md
```diff
@@ -29,38 +30,31 @@ _Diagram of a chain signature in NEAR_
...
-
-2.
-
-3.
+<hr class="subsection" />
 
 ### 2. Signature Request
 
-Once a payload is created and ready to sign, a signature request is made by calling `sign` the deployed smart contract `multichain.near`. This method takes two parameters:
-  - **payload:** A payload (transaction, message, data, etc.) signed by a NEAR account
```
thisisjoshford: The sender must first sign the payload before the next step.

docs/1.concepts/abstraction/chain-signatures.md
gagdiez: I thought that the key is derived from `accountId` and `path`, and the account keys are not involved in that

docs/1.concepts/abstraction/chain-signatures.md
gagdiez: while "payload" is technically correct, I think that transaction is easier to be understood. Thought, we can switch it back if we are going for correctness

docs/1.concepts/abstraction/chain-signatures.md
gagdiez: for what I saw in the code the NEAR account does not need to sign the payload, it is sent raw, and the MPC network signs it

docs/1.concepts/abstraction/chain-signatures.md
thisisjoshford: Correct, but how the new key is created wasn't what I was trying to convey to the reader. I was trying to articulate that "All this is possible (many cross-chain accounts controlled by 1 NEAR account) because of our access keys & account model (that allows for many keys for one account).

Perhaps the second sentence should be:

`Chain Signatures utilizes this ability to generate new keys associated with the NEAR account and sign payloads compatible with other blockchain platforms.`

I think a visual would be helpful here, happy to work on one :)

docs/1.concepts/abstraction/chain-signatures.md
thisisjoshford: Yeah, certainly a fine line to dance here between "technically" vs "easier to comprehend". 
@mattlockyer @kcole16 wdyt is better for EthDevs? Is payload confusing? Should we just use transactions for simplicity / better understanding?

docs/1.concepts/abstraction/chain-signatures.md
thisisjoshford: Hmmm, I could have sworn I heard "The payload needs to be first signed by the NEAR account and then again by MPC" and it stuck in my head as an interesting notion of twice signed transaction. I probably misunderstood. 

@DavidM-D @volovyks can you confirm that I'm mistaken here?

docs/1.concepts/abstraction/mpc.md
```diff
@@ -0,0 +1,37 @@
...
+- They are doing it by using user key shares derived from their root key shares.
+- A root key is never reconstructed, but protocol allows to create signatures using it’s shares.
+
+:::info
+Using MPC, the root key is never reconstructed and it’s never available. User key is never reconstructed as well.
+:::
+
+## How MPC creates a new key
+
+- Once MPC account verification is complete, a root key becomes available to sign a new signature that creates a new key
```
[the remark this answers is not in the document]
gagdiez: this file was a first draft that ended up being orphan (it is in the docs, but linked nowhere)... will delete it in a future PR

PR #1730, opened 2024-02-13
Added js examples for cross contract calls (batch actions)

garikbesson:


garikbesson:
Needs to change repos links after merging new examples [here](https://github.com/near-examples/cross-contract-calls/pull/1).

gagdiez (approved):

PR #1733, opened 2024-02-14
Chain Abstraction

gagdiez:
Creating the section "Chain Abstraction" on docs.

P.S. Tomorrow I'll add figures to it


gagdiez:
@thisisjoshford @bucanero maybe we want to drop the `Chain Abstraction: The holistic view`, since it reads more like a blogposts than an intro to technical docs.

I could also move it to "Learn" ... or simply leave it there.

Opinions?

bucanero:
> @thisisjoshford @bucanero maybe we want to drop the `Chain Abstraction: The holistic view`, since it reads more like a blogposts than an intro to technical docs.
> 
> I could also move it to "Learn" ... or simply leave it there.
> 
> Opinions?

Yes, I'd place this in the "Learn" section, I think it fits there. I also agree the "holistic view" feels more like content for a blog post, so I'd keep it out.

bucanero (approved):

PR #1774, opened 2024-02-29
Migrate Pagoda docs

bucanero:


gagdiez:
I think the RPC particularly could get us in trouble, since we are linking all the other projects to their docs, and Pagoda RPC gets all the attention with its own section. We need to make a decision about this.


`Alerts` and `Enhance API` can go in tools, I'll move them

bucanero:
@thisisjoshford, what do you think about Pagoda RPC docs?
@gagdiez  we can discuss and define together what to do during our docs call.

pkudinov:
Looks good to me. We don't need to promote these tools because they will be discontinued soon, so we don't have to make them more visible on the side menu.

pkudinov (approved):

PR #1786, opened 2024-03-06
add rating component

bucanero:


bucanero:
@thisisjoshford here's an alternative color scheme: https://slack-files.com/T01QNL8F8JE-F06NRRA6E1X-750de99cc4

img

PR #1818, opened 2024-03-13
Update DA light client docs

bucanero:


docs/data-availability/light-client.md
```diff
@@ -15,6 +38,69 @@ A blob submission can be verified by:
...
+
+#### Circuits
+
+Below are the current circuits for the ZK light client.
+
+:::note
+The current circuits serve only a "one-shot" command style for syncing/verification and no autonomous proving.
+:::
+
+Since the ZK client integrates with a Solidity Contract on chain and the circuit must have a statically aligned size, we have to minimise as much calldata as possible, opting to witness verification in the circuit rather than store information.
```
bucanero, line 84:
```suggestion
Since the ZK client integrates with a Solidity Contract on chain and the circuit must have a statically aligned size, we have to minimize as much call data as possible, opting to witness verification in the circuit rather than store information.
```

bucanero, line 84:
```suggestion
Since the ZK client integrates with a Solidity Contract on chain and the circuit must have a statically aligned size, we have to minimize as much call data as possible, opting to witness verification in the circuit rather than store information.
```

docs/data-availability/light-client.md
```diff
@@ -3,8 +3,31 @@ id: light-client
 title: Light Client
 ---
 
+# NEAR light client
+
 A trustless off-chain light client for NEAR with DA-enabled features, such as KZG commitments, Reed-Solomon erasure coding & storage connectors.
 
+:::note
+The NEAR Light client was inspired by the [Helios light client](https://github.com/a16z/helios), where there were once many light client implementations for Ethereum until it was released. 
+:::
```
dndll, line 12: I'd say we probably don't need this in the docs, we can keep it in the README.

dndll (approved):
LGTM as a first intro

docs/data-availability/light-client.md
bucanero, line 12:
```suggestion
```

bucanero: 👍 removed

bucanero:
> LGTM as a first intro. As a side question, how much of an effort would it be to just track the repos' readmes? This would mean there would be less maintenance on this repo for documentation.

Perhaps it's possible, but ideally our docs will go beyond the basic content presented on a repository ReadMe.
For these new features we had to rush things a bit (so docs could be ready for events like ETH Denver), but the plan is to keep iterating so the content grows and provides more information.

PR #1819, opened 2024-03-15
Add example about multi chain relayer + gas station

bucanero:


docs/2.develop/relayers/relayer-gas-example.md
```diff
@@ -0,0 +1,152 @@
+---
+id: relayer-gas-example
+title: Multi-chain Relayer and Gas Station example
```
ravigautham, line 3: Lets update to "Multichain" so that we are consistent everywhere

bucanero, line 3:
```suggestion
title: Multichain Relayer and Gas Station example
```

bucanero: fixed 👍

ravigautham:
@anthony-near @bucanero I think we need to add details on Paymaster accounts to this example as well as details on manual settlement. Since those are both key aspects of a partner integrating with the service. 

Here are some notes that I drafted on Paymasters, but I'm not sure if this should be added to the example or if it should be a separate section? Maybe we should create a separate section in the docs titled "Paymasters and Settlement". Thoughts on this?

Paymaster
- A paymaster represents an address on a destination chain that holds a balance of that chain’s native gas token.
- User addresses on destination chains will be funded directly from paymaster accounts
- Partners that want to integrate with the Multichain Gas Relayer service will need to create, fund, and manage paymaster accounts on the destination chains that they want to have support for
- Manual settlement between the NEAR Gas Station contract and paymaster accounts will also be required on a regular basis to ensure a consistent balance of funds

bucanero:
@ravigautham , if I understand correctly:
- Paymasters are needed if you want to integrate with Multichain Gas Relayer, so perhaps we can add the info directly on the relayer article: https://docs.near.org/develop/relayers/multichain-server
- Manual settlement is related to the gas station contract; then we could add it to the gas station contract article: https://docs.near.org/develop/relayers/gas-station

Then in the Example doc, we could just add a "tip" or "note" with a link to the other articles.

ravigautham:
@bucanero I think that split up makes sense. Let's go with that approach. 

Also in the the "Multichain Relayer Server" docs we should update the supported chains section to the following:

- BSC Testnet
- BSC Mainnet (March 27)
- Ethereum Mainnet (March 27)
- More chains coming soon!

We can remove Solana for now, it will be something we look to add support for in Q2. Or we can also add it under the coming soon section along with Ethereum L2s

bucanero:
Thanks for the feedback @ravigautham , I'll start working on it and ping back when a new draft is ready 👍🏻

ravigautham:
@bucanero Thanks!

Here's the info re-Settlement that Anthony drafted:

Settlement is needed because the Gas Station contract is accumulating NEAR, while the paymaster accounts on foreign chains are spending native foreign chain gas tokens (ETH, BNB, SOL, etc). Manual Settlement involves several steps:

Withdrawing the NEAR held in the gas station contract and swapping for a token that can be bridged
i. may be the native gas token of the foreign chain, another token like USDC that has wide bridge support, or NEAR
Bridging the token from NEAR to the foreign chain
i. Here's an overview of bridging related to NEAR: https://knotty-marsupial-f6d.notion.site/NEAR-Bridging-Guides-f4359bd35c794dc184b098f7ed00c4ce
Sending the native gas tokens to the paymaster accounts on the foreign chains
i. A swap from the bridged token to the native gas token before sending to the paymaster accounts is necessary if the token that was bridged was not the foreign chain native gas token

docs/2.develop/relayers/relayer-gas-example.md
```diff
@@ -0,0 +1,152 @@
...
+
+:::tip
+
+Find the Gas Station Event indexer source code in [this GitHub repository](https://github.com/near/gas-station-event-indexer).
+
+:::
+
+## Running tests
+
+1. Set the transaction details of the EVM transaction you want to send in [`generate_rlp_evm_txn.py`](https://github.com/near/multichain-relayer-server/blob/5b040611f2dc6c6b405b5ec00d5102e3cc27a65c/integration_tests/generate_rlp_evm_txn.py), run the script, and save the RLP hex string output.
```
anthony-near, line 76: The gas station contract was updated to mainly support EIP-1559 transactions, so the note should just say `(Python and Rust output different hex RLP encoded transactions)` and the Rust and Python examples to generate transactions should use the [generate_eip1559_rlp_hex()](https://github.com/near/multichain-relayer-server/blob/feature/MVP_ethers/tests/tests.rs#L24) and [generate_rlp_encoded_transaction(is_eip_1559=true)](https://github.com/near/multichain-relayer-server/blob/feature/MVP_ethers/integration_tests/generate_rlp_evm_txn.py#L7) functions respectively.

anthony-near (changes requested):
Everything looks good except generating eip-1559 transactions

docs/2.develop/relayers/relayer-gas-example.md
bucanero: 👍 note updated

anthony-near (approved):

PR #1868, opened 2024-04-02
Compressed DA into a single page

gagdiez:
DA right now is a lot of pages with very little information on it. More importantly, many are now outdated. I would recommend to have a single page in our docs, and help the solutions team to re-organize the README.md files on the official repo to have the right diagrams in the right places


bucanero (commented):
If we merge the content I think we should also remove the old files, so we don't end up with orphan/abandoned pages

Issue #1961, opened 2024-05-08
Add various versions of MPC contracts


mikedotexe:
**Is your feature request related to a problem? Please describe.**

As a user, I would like to discover the different MPC smart contract addresses in documentation instead of Telegram.

**Describe the solution you'd like**

Add the contract addresses and information to docs, covered in the final message here:

<img width="797" alt="chain-siggy-chaos" src="https://github.com/near/docs/assets/1042667/34e87177-24d4-4e90-93be-51bd5390cf3e">

**Additional context**

Chain signatures is live and I believe we need internal folks to be updating the documentation when there are changes that affect partners and builders. It's unusual to not have an issue created when this is a known blocker for developers.

mikedotexe:
I appears that the testnet contract linked in docs (`multichain-testnet-2.testnet `) here:
https://docs.near.org/build/chain-abstraction/chain-signatures#create-a-chain-signature

It seems that this contract has been down/offline for over a week, with this transaction being the last successful one:
https://testnet.nearblocks.io/txns/6dJzPYBSPgidGvpt7JRX5qeMAjd7p4sMtHXZFanDtfxi

In terms of the `v2` contract, I created this issue after seeing a problem:
https://github.com/near/mpc-recovery/issues/590

I believe the `v5` version is the one we're supposed to use. I've seen this error still:
https://github.com/near/mpc-recovery/issues/591

but it seems like when I immediately try again after a failed transaction times out, it will often work. So I believe it makes sense to change the address to that one.

mikedotexe:
Okay, looks like the highlighted screenshot isn't correct, according to this comment: https://github.com/near/mpc-recovery/issues/513#issuecomment-2117098099

it looks like we need to be pointing to `v2.multichain-mpc.testnet`. I've updated the pull request

gagdiez:
merged on #1975

mikedotexe:
Can we add the v5 one? It seems that people are mentioning using that in the Telegram channels. Can we add details on what that is, where the contract is, how it's different, when to use it, etc.?

I wish I knew the answers so I could modify my PR, but I don't know and it doesn't seem to have been communicated externally as well as internally.

volovyks:
@gagdiez can we add more info about our envs? You can find more information in the comment here: https://github.com/near/mpc-recovery/issues/601

PiVortex:
Fixed by #2035

PR #1975, opened 2024-05-16
Update testnet MPC contract to `v5.multichain-mpc-dev.testnet`

mikedotexe:
I believe this is the one we want to point folks to. See context in #1961


gagdiez (approved):

volovyks:
@gagdiez can we reopen this and add more info about our envs? You can find more information in the comment here: https://github.com/near/mpc-recovery/issues/601

PR #1997, opened 2024-05-30
Update design-system (add DIG components)

bucanero:


gagdiez:
We could consider surrounding them with the <WidgetEditor /> so they become interactive 
- https://github.com/near/docs/blob/master/docs/2.build/3.near-components/what-is.md?plain=1

bucanero:
> We could consider surrounding them with the `<WidgetEditor />` so they become interactive
> 
>     * https://github.com/near/docs/blob/master/docs/2.build/3.near-components/what-is.md?plain=1

great idea, I'll update the PR and add the interactive part so users get a better idea of the components available.

gagdiez (approved):

PR #2031, opened 2024-06-07
chore: minor copy updates that I hope provide more clarity

charleslavon:


bucanero (approved):

docs/2.build/1.chain-abstraction/wallet.md
```diff
@@ -40,7 +40,7 @@ Parsed BTC: send 100 BTC to Attacker
 
 The user would approve the `SOL` transaction but the attacker would also get the `BTC` transaction.
 
-This can be solved by having different keys for any chains that you can't prove could have ambiguous transactions. This means that while an attacker may create ambiguous transactions, it will only be for wallets without assets on the target chain.
+This can be solved by having different keys for any chains that could have ambiguous transactions. This means that while an attacker may create ambiguous transactions, it will only be for wallets without assets on the target chain.
```
charleslavon, line 43: @gagdiez does this update make this sentence more clear?

gagdiez: I think we could just say "having different keys for different chains. That way, even if an attacker makes you sign a transaction that is also valid in another chain, there are no assets to be compromised"

Since trying to talk about ambiguos transactions might create more confusion than anything else

gagdiez:
```suggestion
This can be solved by having different keys for different chains. That way, even if an attacker makes you sign a transaction that is also valid in a foreign chain, there are no assets to be compromised
```

bucanero:
Oops, I thought this PR was ready to merge; I missed that it was still being discussed 😅

charleslavon:
No worries I've added it [here](https://github.com/near/docs/pull/2033)

PR #2037, opened 2024-06-10
WIP: Remove links without a parent

gagdiez:


gagdiez:
@victorchimakanu the docs are not building, apparently some files are importing the content of the files that were deleted. Lets move the content to wherever they are imported, so we can delete the files

bucanero:
> @victorchimakanu the docs are not building, apparently some files are importing the content of the files that were deleted. Lets move the content to wherever they are imported, so we can delete the files

Right, I think these "orphan" files are only imported in this article:
https://docs.near.org/build/primitives/dao

Importing the content into the main article `.md` file should work fine

PR #2077, opened 2024-06-24
Update Rust NFT zero-to-hero tutorials

bucanero:


gagdiez (approved):

gagdiez:
we will be merging `near-cli` into `near-cli-rs` very soon (hopefully tomorrow). This means that `near-cli` commands will work on `near-cli-rs`.

Since the `near-cli` commands are shorter, I wonder if we prefer to leave those, instead of using the verbose form of `near-cli-rs` commands. 

We of course can always change them to "verbose" mode, and once the CLIs are merged, change it back to the "simple mode"

PR #2145, opened 2024-07-15
Chain deployed UI project documentation

bucanero:


docs/2.build/4.web3-apps/chain-ui.md
```diff
@@ -0,0 +1,77 @@
...
+
+:::tip
+
+You can still deploy your own file storage smart contract if you want to have additional customization or if you want total control over your data.
+
+:::
+
+### Pros
+
+- Decentralized front-ends are less susceptible to single points of failure, making them more resistant to attacks and server downtimes
```
andy-haynes, line 25: I think this glosses over the fact that our solution is entirely contingent on hosting a web service that is vulnerable to these exact things 😅

docs/2.build/4.web3-apps/chain-ui.md
```diff
@@ -0,0 +1,77 @@
...
+
+### RPC
+
+An [RPC provider](../../5.api/rpc/providers.md) services requests for chain data. For resilience, gateways should ideally be capable of falling back to a different provider in the event the primary provider is experiencing degraded service.
+
+## Considerations
+
+Compared to classic Web2 hosting solutions, the cost structure of a chain-deployed dApp is very different.
+
+When deploying on chain, you pay for the storage space of your compressed bundles. This means that it's critical to pay attention to the size of dependencies you include in your application, as packages that are bloated or do not support proper tree-shaking will directly affect your cost.
```
andy-haynes, line 58: > pay for the storage space of your compressed bundles

This would be a good place to call out that this a one time cost 😎

docs/2.build/4.web3-apps/chain-ui.md
```diff
@@ -0,0 +1,77 @@
+---
+id: chain-deployed-ui
+title: Chain Deployed UI
+sidebar_label: Chain Deployed UI
+---
+
+[Chain Deployed UI](https://github.com/near/chain-deployed-ui) is a [decentralized frontend solution](frontend.md#decentralized-frontend-solutions) that lets you to deploy dApp frontends directly on the NEAR blockchain. A frontend can be built with typical Web2 tooling (e.g. React + Vite + node.js dependencies), and the resulting compressed bundle can be stored in the state of a smart contract.
```
andy-haynes, line 7: Would this be a good place to call out that this is pre-alpha/experimental? 😅

andy-haynes (changes requested):
Some comments/questions inline but otherwise looking good 😎 

Thanks for putting this together!

docs/2.build/4.web3-apps/chain-ui.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Overview
+
+Thanks to the fact that storage on NEAR is cheap, it's feasible to host small applications on chain, especially if some optimization is added to reduce redundant code between deployed dApps. This offers a straightforward path to decentralized hosting as an alternative to static deployment platforms such as Vercel, GitHub Pages, Render, and others.
```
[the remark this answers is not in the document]
bucanero: Fixed, went with your suggestion:
```
Thanks to the fact that storage on NEAR is cheap, it's feasible to host lightweight applications with optimized bundles on chain.
```

docs/2.build/4.web3-apps/chain-ui.md
```diff
@@ -0,0 +1,77 @@
...
+
+The Chain Deployed UI tooling provides the boilerplate app code (React, Vue, TypeScript), along with the `FileStore` smart contract and basic CDN tools so you can build a dApp, pack it, and store it on chain.
```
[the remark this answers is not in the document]
bucanero: At least from @pavelisnear 's explanation, it looked like the idea is that we are going to offer a smart contract deployed on `mainnet` so users can just upload content to it (as long as they have NEAR tokens to pay for storage)

If the smart contract is not part of the solution, then yes let's move it away.

docs/2.build/4.web3-apps/chain-ui.md
bucanero: Good call. Changed to:

```
When deploying on chain, you pay a one-time fee for the storage space of your compressed bundles.
```

docs/2.build/4.web3-apps/chain-ui.md
bucanero: Right, unless they have a setup with multiple gateways and RPC servers, they could still have a single point of failure.

I'm removing this point for now.

docs/2.build/4.web3-apps/chain-ui.md
bucanero: From our recent chat, we want to clarify with @mpeterdev about how we want to communicate the "Chain deployed UI" readiness.

Since it's in a pre-alpha /experimental stage, I think we should keep this documentation as a draft, and when we think it's “ready to go” then we can publish the docs.

I'm adding a note to this article, even if we keep it as a draft.

frol:
@bucanero Sorry for jumping in, but I'd like to invite you to participate in the [Race of Sloths](https://race-of-sloths.com) - a fun and gamified open-source contributions program. Consider mentioning @race-of-sloths user in your github comment or PR description to join!

See how the flow works here: https://github.com/near/nearcore/pull/11778

andy-haynes (changes requested):
Awesome work @bucanero !

Just a few minor corrections inline + we can use the new URL now 🎉  I've updated the references in the other repo.

Since there is a lot of overlap with the code repo's documentation, it would be nice to remove what's covered here and link to the tutorial document as the primary source for general usage. Not urgent, just something to save us some time as the interface matures.

PR #2153, opened 2024-07-17
Update bounty.md

PiVortex:
Change bug bounty link


bucanero (changes requested):
Having a 2 line article feels wrong. If we don't have content for a "bounty program" article, then let's just add this information to the main higher-level article and remove this file.

gagdiez (approved):

PR #2155, opened 2024-07-18
Publish blog post about 2024-07-11 near.org outage

ewiner:
Added a new blog post about the DNS outage, and updated the announcement bar to be less 📣 about it:

image


website/docusaurus.config.js
```diff
@@ -90,7 +90,7 @@ const config = {
     announcementBar: {
       id: 'id-0004',
```
gagdiez, line 91:
```suggestion
      id: 'id-0005',
```

gagdiez (approved):

PR #2157, opened 2024-07-19
Relayer nonce collisions

SurgeCode:
Added a small section to the relayer page about how to handle high throughput relaying

### DIFF
--- a/docs/2.build/1.chain-abstraction/meta-transactions.md
+++ b/docs/2.build/1.chain-abstraction/meta-transactions.md
@@ -110,7 +110,7 @@ Progress is being made to make this possible in the future.
 
 When running a relayer that handles a large number of transactions, you will quickly run into a `nonce` collision problem. At the protocol level, transactions have a unique number that identifies them (nonce) that helps to mitigate reply attacks. Each key on an account has its own nonce, and the nonce is expected to increase with each signature the key creates.
 
-When multiple transactions are sent from the same access key simultaneously, their nonces might collide. Imagine the relayer creates 3 transactions `Tx1`, `Tx2`, `Tx3` and send them in very short distance from each other, and lets assume that Tx3 has the largest nonce. If Tx3 ends up being processed before Tx1 (because of network delay, or simply because a node picks Tx3 before Tx1), then Tx3 will execute, but Tx1 and Tx2 will fail, because they have smaller nonce!
+When multiple transactions are sent from the same access key simultaneously, their nonces might collide. Imagine the relayer creates 3 transactions `Tx1`, `Tx2`, `Tx3` and send them in very short distance from each other, and lets assume that `Tx3` has the largest nonce. If `Tx3` ends up being processed before `Tx1` (because of network delays, or a node picks `Tx3` first), then `Tx3` will execute, but `Tx1` and `Tx2` will fail, because they have smaller nonce!
 
 One way to mitigate this is to sign each transaction with a different key. Adding multiple full access keys to the NEAR account used for relaying (up to 20 keys can make a significant difference) will help. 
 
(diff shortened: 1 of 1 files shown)


docs/2.build/1.chain-abstraction/meta-transactions.md
```diff
@@ -106,6 +106,21 @@ Progress is being made to make this possible in the future.
 
 </details>
 
+### High volume parallel processing
+
+When running a relayer that handles a large number of transactions, issues will quickly arise due to each transaction requiring a unique nonce. If multiple transactions are sent from the same access key simultaneously, the process becomes error-prone.
+
+Fortunately, this is easy to resolve. Adding multiple full access keys to the NEAR account used for relaying (up to 20 keys can make a significant difference) will help. Then, relaying transactions using instances of the Account object created from the various keypairs associated with the account.
+
+To add a key, generate a new keypair and use the original account object to call the addKey method
```
bucanero, line 115:
```suggestion
To add a key, generate a new key pair and use the original account object to call the `addKey` method:
```

docs/2.build/1.chain-abstraction/meta-transactions.md
```diff
@@ -106,6 +106,21 @@ Progress is being made to make this possible in the future.
...
+
+```typescript
```
bucanero, line 117:
```suggestion
```js
```

gagdiez:
I think what we actually need to explain goes more like this:


At the protocol level, transactions have a unique number that identifies them (`nonce`) that helps to mitigate `reply attacks`. Each key on an account has its own `nonce`, and the `nonce` is expected to increase with each signature the key creates.

When the relayer account (e.g. relayer.near) starts relaying a lot of transactions, it will quickly start to happen that `nonce`s collide. For example, if 3 users interact with the relayer at the same time, the relayer will create 3 transactions `Tx1, Tx2, Tx3` and send them in very short distance from each other. In our case, we can assume that `Tx3` has the largest `nonce`, and `Tx1` the smallest. 

If `Tx3` ends up being processed before `Tx1` (because of some network issues one takes longer to be sent to the RPC, or simply because a node picks `Tx3` before `Tx1`), then `Tx3` will execute, but `Tx1` and `Tx2` will fail, because they have smaller `nonce`!

One way to mitigate this is to sign each transaction with a different key. Since `nonce` are key dependent, we can simply create 20 different keys for `relayer.near`, and go rotating which one we use to sign the transactions. This way, we can 20 transactions created simultaneously without the fear of them colliding on the nonce.

DavidM-D:
Can we just have "PAIN" in bold [linking to this comment](https://github.com/near/NEPs/pull/522#pullrequestreview-1810198990)?

gagdiez (approved):

docs/2.build/1.chain-abstraction/meta-transactions.md
```diff
@@ -106,6 +106,26 @@ Progress is being made to make this possible in the future.
 
 </details>
 
+### High volume parallel processing
+
+When running a relayer that handles a large number of transactions, you will quickly run into a `nonce` collision problem. At the protocol level, transactions have a unique number that identifies them (nonce) that helps to mitigate reply attacks. Each key on an account has its own nonce, and the nonce is expected to increase with each signature the key creates.
+
+When multiple transactions are sent from the same access key simultaneously, their nonces might collide. Imagine the relayer creates 3 transactions `Tx1`, `Tx2`, `Tx3` and send them in very short distance from each other, and lets assume that Tx3 has the largest nonce. If Tx3 ends up being processed before Tx1 (because of network delay, or simply because a node picks Tx3 before Tx1), then Tx3 will execute, but Tx1 and Tx2 will fail, because they have smaller nonce!
```
gagdiez, line 113:
```suggestion
When multiple transactions are sent from the same access key simultaneously, their nonces might collide. Imagine the relayer creates 3 transactions `Tx1`, `Tx2`, `Tx3` and send them in very short distance from each other, and lets assume that `Tx3` has the largest nonce. If `Tx3` ends up being processed before `Tx1` (because of network delays, or a node picks `Tx3` first), then `Tx3` will execute, but `Tx1` and `Tx2` will fail, because they have smaller nonce!
```

gagdiez (approved):

PR #2194, opened 2024-08-01
document the new `ReceiptValidationError::ReceiptSizeExceeded`

jancionear:
Add information about the new `ReceiptValidationError::ReceiptSizeExceeded` error added in https://github.com/near/nearcore/pull/11492.

I'm new to this repo, I looked at `ReceiptValidationError::InvalidDataReceiverId` and put information about my error in the same places.

I also have a longer description of the error that users could find more useful, but I'm not sure where to put it. Can I replace the comment saying "/// Receipt is bigger than the limit."? Or should these docs stay in sync with the source files?

Longer description:

> `ReceiptSizeExceeded` means that there was a receipt above the size limit (currently 4MiB). NEAR will refuse to execute receipts that are above the size limit. The most likely source of such receipts would be cross-contract calls with a lot of large actions (contract deployment, function call with large args, etc).
This error means that the user has to adjust their contract to generate smaller receipts.

Refs: [zulip conversation](https://near.zulipchat.com/#narrow/stream/308695-nearone.2Fprivate/topic/partner.20request.20for.20rpc/near/455717072)

### DIFF
--- a/docs/6.integrations/errors/error-implementation.md
+++ b/docs/6.integrations/errors/error-implementation.md
@@ -442,6 +442,11 @@ pub enum ReceiptValidationError {
     /// An error occurred while validating actions of an ActionReceipt.
     ActionsValidation(ActionsValidationError),
     /// Receipt is bigger than the limit.
+    /// ReceiptSizeExceeded means that there was a receipt above the size limit (currently 4MiB).
+    /// NEAR will refuse to execute receipts that are above the size limit.
+    /// The most likely source of such receipts would be cross-contract calls with a lot of large actions
+    /// (contract deployment, function call with large args, etc).
+    /// This error means that the user has to adjust their contract to generate smaller receipts.
     ReceiptSizeExceeded { size: u64, limit: u64 },
 }
 ```
--- a/docs/6.integrations/errors/error-implementation.md
+++ b/docs/6.integrations/errors/error-implementation.md
@@ -441,6 +441,8 @@ pub enum ReceiptValidationError {
     NumberInputDataDependenciesExceeded { number_of_input_data_dependencies: u64, limit: u64 },
     /// An error occurred while validating actions of an ActionReceipt.
     ActionsValidation(ActionsValidationError),
+    /// Receipt is bigger than the limit.
+    ReceiptSizeExceeded { size: u64, limit: u64 },
 }
 ```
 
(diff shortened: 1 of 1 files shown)


telezhnaya (commented):
Let's also add some comments to this page? https://docs.near.org/api/rpc/transactions

Also, we need to give some guidelines what should the user do in this case
(Contact the contract owner?)

jancionear:
> Let's also add some comments to this page? https://docs.near.org/api/rpc/transactions
> 
> Also, we need to give some guidelines what should the user do in this case (Contact the contract owner?)

I'm not sure if it deserves to be on the transactions page. It's a low level error, I don't see anything about other `ReceiptValidationErrors` in there. Even basic errors like `GasLimitExceeded` aren't there.

frol:
@jancionear Sorry for jumping in, but I'd like to invite you to participate in the [Race of Sloths](https://race-of-sloths.com) - a fun and gamified open-source contributions program. Consider mentioning @race-of-sloths user in your github comment or PR description to join!

See how the flow works here: https://github.com/near/nearcore/pull/11778

telezhnaya:
My comment is inspired by https://github.com/near/docs/commit/fef1566eec69499eaa23faa993345b13ae771680
I think we need to add at least any comment how should it be handled correctly

jancionear:
> My comment is inspired by [fef1566](https://github.com/near/docs/commit/fef1566eec69499eaa23faa993345b13ae771680)

I think there's a big difference between `ReceiptSizeExceeded` and `ShardCongested`. `ReceiptSizeExceeded` belongs to a common category of errors that occur when receipt execution fails because of something. Rpc users already know how to handle errors like this.
`ShardCongested` is a whole new category on its own, it doesn't belong to any of the existing categories, so it makes sense that it has a special place in the transactions doc. It's a special case that rpc users now need to take into account.

I agree that it'd be nice to have a detailed description of the error somewhere, but I'm hesitant to put it on the main transactions page.

telezhnaya (approved):
I am fine without touching transactions page, and feel free to merge it as it is, but I still think it's better to leave additional comment here, right where you already added some lines, with the comment how should the user handle this

jancionear:
> but I still think it's better to leave additional comment here, right where you already added some lines, with the comment how should the user handle this

I added the longer comment under the existing short one.

PR #2243, opened 2024-10-03
add zk page

PiVortex:
Adding a page about ZK to primitives

### DIFF
--- a/docs/2.build/5.primitives/zk.md
+++ b/docs/2.build/5.primitives/zk.md
@@ -1,0 +1,43 @@
+---
+id: zk
+title: Zero Knowledge Proofs
+sidebar_label: Zero Knowledge Proofs
+hide_table_of_contents: false
+---
+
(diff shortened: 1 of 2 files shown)


gagdiez:
@PiVortex generally our primitive pages are all about how to exactly use the primitive, do you think we can add code snippets here?

PiVortex:
Since we are not a maintainer of these projects I think it would bring in too much upkeep to update code snippets when the code changes. Furthermore, we would likely just be copying and pasting code that is in the link that we are guiding them to. For things like FTs and NFTs because they are purely NEAR native and the docs is the main source of knowledge for using these of course we have the node snippets for them but I think some sections we shouldn't include code. We could ask projects to add to these docs and upkeep them if they want to be featured but I reckon they would rather just have a link referencing their own docs.

It's important to reference what is available in the docs but maybe this is then suited for somewhere else. I would also like to add more oracle solutions to that page but again I wouldn't want to bring code snippets in unless necessary, more comparing the different solutions and what they offer and linking to the relevant docs.

gagdiez:
This has shown no progress for a month, maybe we close it @PiVortex ?

PR #2244, opened 2024-10-04
feat: add eth wallets docs and blogpost

gagdiez:


frol:
@race-of-sloths score 13

PR #2251, opened 2024-10-09
NEAR API section (ex. JavaScript API)

garikbesson:


bucanero (approved):

PiVortex:
This PR helps #2321 but does not include the Python API and how to sign a message as the APIs do not implement a specific method to do this

garikbesson:
@PiVortex amazing work, thank you! It would be nice to have Response Examples for each example as well, but we can create another issue for that

Since near-api-examples are migrated to proper organization, let's change code snippets URLs here. After that I think PR can be merged

gagdiez (approved):

thisisjoshford:
fixes #2321

PR #2318, opened 2024-11-20
words about key value storage

PolyProgrammist:
Currently there is no page that explains kay-value storage in context of near sdk. Adding the information here


PolyProgrammist:
Moved here https://github.com/near/docs/pull/2325

PR #2319, opened 2024-11-20
make external calls more clear for rust

PolyProgrammist:
it's important to emphasize `ext_contract` macro which is difficult to find here


PolyProgrammist:
Moved here https://github.com/near/docs/pull/2325

PR #2325, opened 2024-11-21
Fixes for rust sdk

PolyProgrammist:


gagdiez:
@PolyProgrammist is this ready to be checked? I see that it is still marked as a `Draft`

website/sidebars.js
```diff
@@ -139,6 +140,7 @@ const sidebar = {
                 "build/smart-contracts/anatomy/serialization-protocols",
                 "build/smart-contracts/anatomy/reduce-size",
                 "build/smart-contracts/anatomy/reproducible-builds",
+                "build/smart-contracts/anatomy/macros",
```
gagdiez, line 143: good call adding this here

PR #2392, opened 2024-12-23
update usages of `cargo-near`

dj8yfo:
follow-up on https://github.com/near/cargo-near/pull/262 in the spirit of #2390


docs/2.build/2.smart-contracts/quickstart.md
```diff
@@ -435,6 +435,6 @@ At the time of this writing, this example works with the following versions:
 - node: `20.18.0`
 - rustc: `1.81.0`
 - near-cli-rs: `0.15.1`
-- cargo-near: `0.10.1`
+- cargo-near: `0.13.2`
```
dj8yfo, line 438: changes in https://github.com/near/docs/pull/2390/files were somewhat inconsistent, by suggesting to use option, but not changing cmd version

docs/3.tutorials/examples/advanced-xcc.md
```diff
@@ -290,9 +290,7 @@ In order to deploy the contract you will need to create a NEAR account.
 Go into the directory containing the smart contract (`cd contract-advanced-ts` or `cd contract-advanced-rs`), build and deploy it:
 
 ```bash
-cargo near build
```
dj8yfo, line 293: it's somewhat pointless to do `cargo near build` in the same sequence with `cargo near deploy`, as the latter will do a build step anyway

docs/3.tutorials/fts/6-marketplace.md
```diff
@@ -42,7 +42,7 @@ Let's start by building both the finished FT contract and the marketplace contra
 
 
 ```bash
-cd market-contract && cargo near build && cd ..
+cd market-contract && cargo near build non-reproducible-wasm && cd ..
```
dj8yfo, line 45: in such occurencies where full command is required as part of a script, `non-reproducible-wasm` variant token was added to prevent prompts to used, 
 
it other places having only `cargo near build` prefix is still correct, as 2 choices will be suggested in interactive menu

gagdiez (approved):

PR #2393, opened 2024-12-23
chore: update `near-cli-rs` versions mentioned

dj8yfo:


bucanero (commented):
Did you test these articles and tutorials using the new CLI version?

gagdiez (approved):

gagdiez:
there was no major change from `0.11` to `0.17`, in the sense that commands did not change, so I would expect everything to still work

PR #2403, opened 2025-01-02
feat: minor edits to introductions & fastauth deprecation

gagdiez:
This is a collection of minor edits to our introduction docs. I have also added a deprecation notice on fastauth.


docs/1.concepts/protocol/access-keys.md
```diff
@@ -2,14 +2,14 @@
 id: access-keys
 title: Access Keys
 ---
-In all blockchains, users control their accounts by holding a `private key` (a secret only they know) and using it to sign [transactions](./transactions.md).
+In all blockchains, users control their accounts by holding a [`private key`](](https://en.wikipedia.org/wiki/Public-key_cryptography) (a secret only they know) and using it to sign [transactions](./transactions.md).
```
bucanero, line 5: Looks the markdown link tag is broken

yarn.lock
bucanero, line 1: I think We should keep the lock file to avoid building issues with JS libraries

gagdiez: the `yarn.lock` is outside the `website` folder, meaning that somebody did `yarn install` by mistake on the root (`/`) instead of installing dependencies for docusaurus (on `/website`)

docs/1.concepts/protocol/access-keys.md
gagdiez: thanks, completely missed that

gagdiez:
```suggestion
In all blockchains, users control their accounts by holding a [`private key`](https://en.wikipedia.org/wiki/Public-key_cryptography) (a secret only they know) and using it to sign [transactions](./transactions.md).
```

yarn.lock
bucanero: Oh, my bad. Thanks for clarifying!

bucanero (approved):

PR #2474, opened 2025-02-14
add clear state tool

bucanero:


flmel (commented):
As I understand this cli tool does:
* deploys the clean-state contract to the given account
* calls the clean fn on the contract deployed

Since we already rely on near-cli maybe consider adding steps how to deploy and use using near-cli alone

flmel (approved):

PR #2478, opened 2025-02-18
fix: update Omni Bridge flowchart

r-near:
Fix the flowchart. Bitcoin part is valid, but others just verify the signature

### DIFF
--- a/docs/chain-abstraction/omnibridge/implementation.md
+++ b/docs/chain-abstraction/omnibridge/implementation.md
@@ -100,13 +100,11 @@ flowchart TD
     
     MPCSign --> Chain{Destination Chain}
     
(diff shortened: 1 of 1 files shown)


bucanero (approved):

bucanero:
@r-near  thanks for the PR 👍 

btw, would you like to be added as `CODEOWNER` for the `/chain-abstraction/omnibridge/` files? let me know 🙏

r-near:
> @r-near thanks for the PR 👍
> 
> btw, would you like to be added as `CODEOWNER` for the `/chain-abstraction/omnibridge/*` files? let me know 🙏

Yes please! You can also add @karim-en

PR #2482, opened 2025-02-19
Move chain abstraction to a single folder

bucanero:


thisisjoshford (dismissed):
I see some broken links still in the build, is this ready for review? Also, do we have redirects in place for all of these links that will be affected with this major migration?

bucanero:
> I see some broken links still in the build, is this ready for review? Also, do we have redirects in place for all of these links that will be affected with this major migration?

yes, it's ready for review, the redirects will be added on Render once the PR is merged.
Also, no broken internal links, just some anchors, but that's unrelated to this PR.

Issue #624, opened 2021-03-01
[DOC] Better explain finality


bowenwang1996:
Currently there does not seem to be a good explanation of how finality works for the general audience (i.e, in addition to technical spec). The docs [here](https://docs.near.org/docs/roles/integrator/faq#how-does-on-chain-transaction-finality-work) is outdated and should be revamped.

doriancrutcher:
Is this in regards to transaction finality or block finality?

PR #2514, opened 2025-04-08
Update info on finalities

gagdiez:
We are missing one finality parameter (`near-final`) which provides more guaranties than `optimistic` while being a block faster than `final`

### DIFF
--- a/docs/api/rpc/setup.md
+++ b/docs/api/rpc/setup.md
@@ -82,6 +82,14 @@ The block IDs of transactions shown in [NearBlocks Explorer](https://testnet.nea
 
 ## Using `finality` param {#using-finality-param}
 
-The `finality` param has two options: `optimistic` and `final`.
-1. `optimistic` uses the latest block recorded on the node that responded to your query _(< 1 second delay after the transaction is submitted)_
-2. `final` is for a block that has been validated on at least 66% of the nodes in the network _(usually takes 2 blocks / approx. 2 second delay)_
+The `finality` param has three options: `optimistic`, `near-final` and `final`:
+
+1. `optimistic` the transaction is in a block that might be skiped _(< 1 second delay after the transaction is submitted)_
+2. `near-final` the transaction is in a block that is irreversible, unless at least one block producer is slashed _(1 block / approx. 1 second delay)_ 
+3. `final` the block is final and irreversible _(usually takes 2 blocks / approx. 2 second delay)_
+
+:::info
+
+The `near-final` finallity has enough guarantees for any normal operations, and thus should be preferred for most applications
+
+:::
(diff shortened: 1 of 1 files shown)


docs/api/rpc/setup.md
```diff
@@ -82,6 +82,14 @@ The block IDs of transactions shown in [NearBlocks Explorer](https://testnet.nea
 
 ## Using `finality` param {#using-finality-param}
 
-The `finality` param has two options: `optimistic` and `final`.
-1. `optimistic` uses the latest block recorded on the node that responded to your query _(< 1 second delay after the transaction is submitted)_
-2. `final` is for a block that has been validated on at least 66% of the nodes in the network _(usually takes 2 blocks / approx. 2 second delay)_
+The `finality` param has three options: `optimistic`, `near-final` and `final`:
+
+1. `optimistic` the transaction is in a block that might be skiped _(< 1 second delay after the transaction is submitted)_
```
gagdiez, line 87:
```suggestion
1. `optimistic` the transaction is in a block that might be skipped _(< 1 second delay after the transaction is submitted)_
```

bucanero (approved):

docs/api/rpc/setup.md
```diff
@@ -82,6 +82,14 @@ The block IDs of transactions shown in [NearBlocks Explorer](https://testnet.nea
...
 ## Using `finality` param {#using-finality-param}
 
-The `finality` param has two options: `optimistic` and `final`.
-1. `optimistic` uses the latest block recorded on the node that responded to your query _(< 1 second delay after the transaction is submitted)_
-2. `final` is for a block that has been validated on at least 66% of the nodes in the network _(usually takes 2 blocks / approx. 2 second delay)_
+The `finality` param has three options: `optimistic`, `near-final` and `final`:
+
+1. `optimistic` the transaction is in a block that might be skipped _(< 1 second delay after the transaction is submitted)_
+2. `near-final` the transaction is in a block that is irreversible, unless at least one block producer is slashed _(1 block / approx. 1 second delay)_ 
+3. `final` the block is final and irreversible _(usually takes 2 blocks / approx. 2 second delay)_
```
frol, line 89: `optimistic` still requires a block to be produced and broadcasted, so in the worst case if transaction landed right after the previous block was produced it will take 1 block to wait until the next block will start forming, then 1 block to get included+converted to receipt in the block (at this point, 2 "seconds" since submission, the transaction will be in an "optimistic" block), then 1 block to execute the receipt + N blocks for cross-account interaction (if needed), and so if someone is looking for "optimistic transaction execution result", they will see it in the block T+3, where T is the optimistic block height at the moment of transaction submission, the block with the transaction result will become "near-final" at T+4 timestamp, and the block will become "final" at T+5 timestamp.

Also, with nearcore 2.6.0 (released yesterday to testnet), the block production time was reduced to 0.6s, so the seconds reference might need to be adjusted accordingly (T+3 will become 1.8s instead of 3s)

gagdiez:
I'll deploy it for now, if further clarification is needed lets open a new issue/pr

This closes #624

PR #2523, opened 2025-05-05
feat: contracts list

### LINKED ISSUE #2260: [DOC] Catalog of Contracts
We need a page that lists common contracts that people might want to use with links to their code and compiled wasm so they can easily deploy them

flmel:
Closes #2260 
Adds a contract list under resources nav 
Screenshot 2025-05-05 at 4 48 14 pm


flmel:
We will add short description to each contract

gagdiez (approved):

PR #2526, opened 2025-05-09
Update document to recommend epoch sync instead of snapshot

walnut-the-cat:
Now with free state snapshot going away, we want to encourage users to use epoch sync. Making changes to reflect that.


walnut-the-cat:
looks like bunch of unrelated broken links are blocking this PR to be merged..?

gagdiez (approved):

PR #2539, opened 2025-05-14
Gleap security update inline

trechriron:
## **This PR is dependent on #2537**  <-- release this second.

* Added two new functions to Root.js; initializeGleap() and isNewTab().
* The new initializeGleap function includes code that addresses the NEAR Gleap security vulnerability.
* This code was tested with this branch rebased on the two previous PRs. The site launches and Gleap functions as expected.
* In our Gleap instance, see the new column Test Messages for messages sent from localhost:3000


trechriron:
Fixed to only include commits/changes to Root.js. See new PR #2540

PR #2541, opened 2025-05-15
Gleap render test - DO NOT MERGE

trechriron:
## **DO NOT MERGE**

This PR is master with rebase of previous 3 PRs to test Render deployment.


trechriron:
The logic in the initializeGleep function was incorrect. This caused navigating back to the home page after redirect to Discord to not load the Gleap side banner. The fix restores the correct behavior.

PR #2576, opened 2025-06-10
Adds "AI Agents" to Not Found page

elliotBraem:
Ran into this because of a broken shade agent link (https://github.com/NearDeFi/shade-agent-template/pull/7), and noticed there is no suggested route to building AI agents.

Just a suggestion; maybe there should be a different ordering or there is better place to start


PiVortex:
That link was just a mistake in the repo as this was the original place i was putting in the docs. It is not used anywhere else. 

We have a section for AI agents 
image

PR #2594, opened 2025-06-21
feat: add section about Global Contracts in docs for `near-api-js`

denbite:


docs/tools/near-api.md
```diff
@@ -1160,6 +1160,88 @@ Unlike many other blockchains, contracts on NEAR are mutable, meaning you have t
...
+
+### Deploy a Global Contract {#deploy-a-global-contract}
+
+If you've ever deployed the same contract code to multiple accounts, you’ve likely noticed that each deployment requires you to pay the full storage cost again — since the size of the WASM file determines how much `NEAR` is locked on the account.
+
+Global Contracts solve this inefficiency by allowing the same contract code to be shared across multiple accounts, so storage cost is paid only once.
+
+There are two ways to reference a global contract:
+- By account — The contract code is tied to another account. If that account later deploys a new version of the contract, your account will automatically start using the updated code — with no need for redeployment.
+- By hash — You reference the contract by its immutable code hash. This ensures you're always using the exact same version, and it will never change unless you explicitly redeploy with a different hash.
```
bucanero:
```suggestion
- **By account:** The contract code is tied to another account. If that account later deploys a new version of the contract, your account will automatically start using the updated code, with no need for redeployment.
- **By hash:** You reference the contract by its immutable code hash. This ensures you're always using the exact same version, and it will never change unless you explicitly redeploy with a different hash.
```

PR #2610, opened 2025-07-02
feat: add a network marquee component for broadcasting multi-chain signatures

### LINKED ISSUE #2577: [DOC] Explicitely state that EVM accounts are natively supported
### Documentation Type

New Documentation

### What needs documentation?

@think-in-universe suggested that we explicitly state `EVM accounts` as a type of account on NEAR Protocol, which is a great idea

### Acceptance Criteria

- [ ] State that we support `evm-like accounts` on https://docs.near.org/protocol/account-model
- [ ] State that we support `evm-like accounts` on https://docs.near.org/protocol/account-id
- [ ] Give a quick look to docs to see if it would be useful to state it anywhere else  

### Resources & References

_No response_

### Priority

🔴 P0 : Missing Critical Documentation

matiasbenary:
I created a marquee component to showcase the benefits of Chain Signatures.
I think I might need to refine the text.

As for this component, I couldn't find another page to add it to, so I included it only on:
http://localhost:3000/protocol/account-model
http://localhost:3000/protocol/account-id
Captura desde 2025-07-01 21-17-38
image

Fixes #2577


gagdiez:
kind of an overkill, lets create something simpler, but this branch can live in our github to maybe be used later

Issue #2566, opened 2025-05-29
[DOC] Update Primitives to use Global Contracts


gagdiez:
### Documentation Type

Update Existing Documentation

### What needs documentation?

Global contracts are a huge win for primitives such as FT, NFT and DAO, as developers will now be able to create their own tokens or decentralized organizations without the need to deploy contracts and pay for their storage (in the case of DAO, this is up to 6N per DAO released)

We need to deploy global FT, NFT and DAO contracts in `testnet` and `mainnet` and update our docs to use them instead of pre-compiled `wasm`

### Acceptance Criteria

- [ ] Deploy reproducible FT contract
- [ ] Update FT section
- [ ] Deploy reproducible NFT contract
- [ ] Update NFT section
- [ ] Deploy reproducible DAO contract
- [ ] Update DAO section

### Priority

🔴 P0 : Missing Critical Documentation

PR #2631, opened 2025-07-15
Update global contract docs

bucanero:
Need to merge the Rust examples , https://github.com/near-examples/near-api-examples/pull/8


garikbesson:
Hi @bucanero, I came across this PR working on this issue: [Update Primitives to use Global Contracts](https://github.com/near/docs/issues/2566).

As a developer, I would like to see in docs how to deal with contract state using Global Contracts? For example, if a global contract with a new update gets a new state structure, how should I update it on my contract deployed by hash? Should I call a migration method manually? Or the migration method will somehow be called automatically since all contracts get updates from a global one.

What do you think will it be worth covering this area?

denbite (approved):
lgtm

PR #2674, opened 2025-08-02
Migrate donation example to token tutorial

iszynero1:
Converted donation example to focus on NEAR token transfers:
- Core token handling methods
- Security checks
- Testing examples


gagdiez:
This PR simply deletes most of the doc without adding any meaningful addition, I am declining it

PR #2675, opened 2025-08-02
Final randomness tutorial formatting

bliss4fx:
Converted example to focus on on-chain randomness:
- Blockhash implementation
- Oracle alternative
- Security considerations
- Testing strategies


gagdiez:
This PR just deletes most of the doc without any meaningful addition:
- it does not explain the code to use randomness based on the current tutorial
- it does not explain testing
- it does not explain security considerations

PR #2676, opened 2025-08-05
Merge docs and tutorials into a single page

gagdiez:
Re-organizing the docs to follow the diataxis framework both within topics as across-topics.

To be more explicit, now the sidebar will be:

```
- Topic 1
  - Concepts
  - Tutorials
  - Reference
- Topic 2
  - Concepts
  - Tutorials
  - Reference
- Topic 3
  - Concepts
  - Tutorials
  - Reference
```

While the top bar navigation will be:

```
- Concepts
  - Topic 1
  - Topic 2
  - Topic 3
- Tutorials
  - Topic 1
  - Topic 2
  - Topic 3
- Reference
  - Topic 1
  - Topic 2
  - Topic 3
```


thisisjoshford (changes requested):
I'm really not a fan of having double menus for "docs" section:

Screenshot 2025-08-05 at 9 44 28 PM

I agree with having tutorials link to the relevant docs section but duplicating this effort here seems off. Also, if you go to RPC or any page away from docs, you cant get back to it without going through the drop down selection menu.

----

Explorer, Data API, Testnet Faucet all have no left nav:

Screenshot 2025-08-05 at 9 48 16 PM

Screenshot 2025-08-05 at 9 49 07 PM

------


Will prob need to adjust color for these icons in dark mode:

Screenshot 2025-08-05 at 9 44 43 PM

________
I feel like Developer Tools should have what is in core libraries as I think of our CLI, SDK, and API dev tools as being part of that. Perhaps we should rename it to Developer Resources or something similar?

Screenshot 2025-08-05 at 10 00 34 PM

________

I'm not sure how I feel about having the CLI in NEAR protocol and API in app development. I think they both should be under NEAR?


Screenshot 2025-08-05 at 10 06 43 PM
Screenshot 2025-08-05 at 9 58 31 PM

We can chat more about all this in our next working session tomorrow @gagdiez

thisisjoshford (approved):

PR #2690, opened 2025-08-12
Add using global contracts option to primitives (ft, nft, dao)

garikbesson:


website/src/components/docs/primitives/dao/near-cli/deploy-using-global-account-id.md
```diff
@@ -0,0 +1,10 @@
+```bash
```
gagdiez: lets put all the commands **within** the respective `.md` files

PR #2694, opened 2025-08-13
Near Drop Tutorial

Festivemena:
Migrated the Near Drop Tutorial into a step by step tutorial that anyone can follow to build an end to end Near Drop integrated Solution with
1. Wallet Integration for Near Accounts
2. Drop Creation Interface
3. Key Generation and Distribution tools
4. QR Code support for easy Sharing
5. Claiming Interface for both new and existing users

With this tutorial, you can now create and claim token drops (Near Drops) following the procedures - no technical knowledge required!

### DIFF
--- a/docs/tutorials/neardrop/access-keys.md
+++ b/docs/tutorials/neardrop/access-keys.md
@@ -1,0 +1,200 @@
+---
+id: access-keys
+title: Access Key Management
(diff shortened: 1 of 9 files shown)


Festivemena:
Fixed branch

PR #2788, opened 2025-09-27
fix: resolve Next button navigation loop in protocol/basics

MaximoCorrea1:
The "Next" navigation button on the "What is NEAR?" page (/protocol/basics) was creating a navigation loop, scrolling back to the same page instead of advancing to the next documentation section.

Root Cause:
The issue was in 'sidebars.js' where 'protocol/basics' was referenced twice:
1. As a category link: link: { type: 'doc', id: 'protocol/basics' }
2. As the first item in the "Getting Started" section

This dual reference caused Docusaurus navigation to create a circular loop.

Solution:
Removed the category link from the NEAR Protocol section in 'sidebars.js'. Now the page only appears once in the navigation hierarchy, allowing proper sequential navigation.

Testing
- [x] Reproduced the issue locally
- [x] Applied the fix by removing the duplicate category link
- [x] Verified "Next" button now correctly navigates to "Create a Testnet Account"
- [x] Verified "Previous" button works correctly on the next page
- [x] Confirmed navigation flow is now working as expected

Files Changed
- 'website/sidebars.js' - Removed duplicate 'protocol/basics' reference

This fix resolves the navigation loop issue reported on the What is NEAR? documentation page.


gagdiez:
changed all `type: doc` to `type: generated-index`, that seamed to fix it, thought it now generates warning of possibly duplicated links

PR #2869, opened 2025-11-10
fix: resolve CSS newsletter issues w/ new and legacy mailchimp forms

thisisjoshford:
New DevNewsletter was displaying dark text on dark background and was unreadable. Legacy mailchimp newsletters had dark text with a whitebackground that was auto-imported in and we had to resolve the two.

### DIFF
--- a/website/src/pages/newsletter.jsx
+++ b/website/src/pages/newsletter.jsx
@@ -12,56 +12,6 @@ const API_BASE = 'https://tmp-docs-ai-service.onrender.com';
 
 const fetcher = (url) => fetch(url).then((res) => res.json());
 
-// Custom CSS sanitization to remove white tones and dark colors for legacy Mailchimp markup.
-const containsWhiteTone = (value) => {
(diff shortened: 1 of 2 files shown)


joe-rlo:
can you just change 'weekly' to 'monthly' at the top?

joe-rlo (approved):

bucanero (approved):

gagdiez (changes requested):
this solution is actually way to complex, since the only thing we need to change is how we are resetting the `css`

luckily, the old mailchimp comes with their own `css` class wrapper, so we can simply **not** reset anything, except the color within the mailchimp wrapper

So a simpler solution is:

1. Do not use `Card` (since it imposes its own color schemas)
2. Set a specific color for the old mailchimp
3. Simply let the new ones use the color schemes of our docs (i.e. do not impose any css changes on them)

gagdiez (approved):

PR #2971, opened 2026-02-17
Fix off-by-one in <File> line-range extraction and add regression test

afurm:
## Summary
`<File ... start="x" end="y" />` snippet extraction had an off-by-one bug that included one extra line after `end`.

This PR:
- fixes the line-range slicing so `end` does not include the next line
- adds a regression test for `start="2" end="3"` to ensure only lines 2–3 are returned
- adds a small `require.main` guard/export so the helper can be tested directly without changing normal script execution

## Before / After
- Before: `start=2, end=3` returned lines `2, 3, 4`
- After: `start=2, end=3` returns lines `2, 3`

## Testing
```bash
node --test website/scripts/copy-md-to-static.test.js

### DIFF
--- a/website/scripts/copy-md-to-static.js
+++ b/website/scripts/copy-md-to-static.js
@@ -54,7 +54,7 @@ function formatCodeBlock(codeData, language = '') {
 
   const codeLines = String(code).split('\n');
   const start = startMatch ? Math.max(parseInt(startMatch[1]) - 1, 0) : 0;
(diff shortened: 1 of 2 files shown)


gagdiez:
hi @afurm , thanks for the PR! could you point which `.md` is showing problems that this PR fixes? I'm trying to figure out if this is a real issue, or the `+1` is actually not affecting the results

PR #2996, opened 2026-04-08
Fix redirect for /protocol/access-keys

### LINKED ISSUE #2995: Broken external link
The external link https://docs.near.org/protocol/access-keys gives an error, we need to find which link is the right one and create a redirect on `docs.json`

irondevrel:
Fixes #2995

This PR adds a redirect in `docs.json` to handle the old URL `/protocol/access-keys` which currently returns a 404 error.

The redirect now points to the correct location: `/protocol/accounts-contracts/access-keys`

**Changes:**
- Added redirect entry in `docs.json` for `/protocol/access-keys` → `/protocol/accounts-contracts/access-keys`


gagdiez (approved):

gagdiez (changes requested):
why did you delete a whole section of the docs.json ❌

PR #3004, opened 2026-04-13
Update/near go sdk update

Emir-Asanov:
Update near go sdk documentation, add examples, add new sections, make go sdk, similar to rust, js, python sdk's.


vlmoon99:
Hi, I want to confirm that it's a part of Near GO SDK contribution, and I already reviewed it and can say that both code and docs good and write in the correct way, if you will have any questions or maybe we need to change something - ping me here and Emir as well we will perform all necessary changes in order to add new Near GO SDK documentation

PR #3031, opened 2026-05-07
Split web login into wallet login and social login pages

### LINKED ISSUE #2911: Web Login: Add new methods and decide on existing ones
When we started to create the `Web Login` section, we wanted to list all possible ways to connect to NEAR, but having more than 3 options might be detrimental for devs - as we fall into The Paradox of Choice.

We might want to cut down our Web Login section to fewer options, and personally I am tilting towards:

- **Privy**: amazing DevX for non-custodial social login
- **Hot Selector**: amazing DevX for custodial wallet login

It is still important to name other possible methods:

- Wallet Selector
- Web3Login
- FastAuth
- Fastintear?

We might actually prefer to divide Web Login into two simple pages: "Social Login" and "Wallet Login", were we showcase the main methods (Privy & Hot Selector) while listing the others (we might need to move the existing docs somewhere else)

matiasbenary:
Close #2911

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -299,6 +299,7 @@
                     "group": "Concepts",
                     "pages": [
                       "web3-apps/concepts/web-login",
+                      "web3-apps/concepts/social-login",
                       "web3-apps/concepts/data-types"
                     ]
                   },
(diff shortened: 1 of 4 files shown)


gagdiez:
I do not feel it is necessary to separate the "concepts" into two files, in any case, we could remove the concept and instead had two tutorials: one for wallet login, one for social login

PR #3050, opened 2026-06-01
chore: add redirects links

matiasbenary:
Redirect links have been added for the old links, in case any pages refer to the documentation


thisisjoshford (approved):
@matiasbenary can you add a description for this PR that explains why we're introducing these redirects?

PR #3079, opened 2026-07-21
docs: add Homebrew install option for NEAR CLI

r-near:
## Summary

- add a Homebrew tab (`brew install near`) to the NEAR CLI installation page, now that the `near` formula is in Homebrew/core (near/devex#56)
- add the same Homebrew tab to the inline install options on the account-creation guide (kept inline so the getting-started flow stays self-contained)
- point the localnet tutorial prerequisite at the installation guide instead of the GitHub repo

## Validation

- `npx --yes mint validate`
- `npx --yes mint broken-links`

### DIFF
--- a/getting-started/create-account.mdx
+++ b/getting-started/create-account.mdx
@@ -47,7 +47,40 @@ Need `testnet` funds? try using one of our [faucets](/getting-started/faucet)
 
 When developing smart contracts you will expend lots of time interacting with the NEAR blockchain through the command line interface (CLI).
 
-First, you will need to install the NEAR CLI, for which multiple installation methods are available (npm, Homebrew, Cargo, and standalone installers) - pick your favorite from the [installation guide](/tools/cli#installation).
+First, you will need to install the [NEAR CLI](/tools/cli#installation):
+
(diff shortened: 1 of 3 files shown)


getting-started/create-account.mdx
```diff
@@ -47,34 +47,7 @@ Need `testnet` funds? try using one of our [faucets](/getting-started/faucet)
...
-  ```
-  </Tab>
-  <Tab title="Windows (binaries)">
-
-  ```bash
-  irm https://github.com/near/near-cli-rs/releases/latest/download/near-cli-rs-installer.ps1 | iex
-  ```
-  </Tab>
-</Tabs>
+First, you will need to install the NEAR CLI, for which multiple installation methods are available (npm, Homebrew, Cargo, and standalone installers) - pick your favorite from the [installation guide](/tools/cli#installation).
```
gagdiez: lets bring the table back, in the quickstart we do not want people (I guess agents now?) switching pages

so this is the only place were it is fine to duplicate data

gagdiez (approved):

